Issue #59, opened 2019-01-12
Searchable Select Boxes


Yanchek99:
Allows select boxes to be searchable to more easily select

PR #60, opened 2019-01-12
#59 searchable select

Yanchek99:
#59 make select fields searchable


Yanchek99:
Retrying with new hound configs

PR #1660, opened 2026-06-16
[codex] Replace workout position fields with drag ordering

### LINKED ISSUE #1638: Replace position fields with drag-and-drop ordering of workout parts
## Goal

Hide `position` from users entirely. Workout parts (top-level exercises and segments, interleaved) are reordered by drag and drop; exercises within a segment are reordered the same way. Position inputs become hidden fields maintained automatically (on drop and on add/remove). New parts append at the end.

Follow-up to #1608, which scoped exercise positions to their workout part.

## Current state

- Editable `Position` number inputs in `app/views/workouts/_exercise_fields.html.slim` and `_segment_fields.html.slim`.
- Top-level exercises and segments share one position sequence per workout; segment exercises restart at 1 (`app/models/concerns/exercise_position_validation.rb`, `app/models/segment.rb`, `Workout#ordered_parts`, `cf/docs/decisions.md` 2026-06-05 entry).
- Partial unique DB indexes: `exercises [position, segment_id] WHERE segment_id IS NOT NULL`, `exercises [position, workout_id] WHERE segment_id IS NULL`, `segments [position, workout_id]`.
- `nested_form_controller.js` assigns max+1 positions on add and dispatches `nested-form:add` / `nested-form:remove` events.
- No drag-drop library in `package.json`.

## Key constraint: the unique-index swap problem

Reordering **persisted** parts breaks today's setup at two layers:

1. **Validations** query the DB and only exclude `self`, so swapping two persisted siblings (A: 1→2, B: 2→1) fails validation — B's row still holds position 2 when A validates.
2. **DB indexes**: Rails autosave updates children one at a time inside the save transaction, so an upward move writes a position still held by a not-yet-updated sibling → `PG::UniqueViolation`. The partial unique indexes cannot be made deferrable (Postgres deferrable constraints don't support `WHERE`); the deferrable equivalent (`EXCLUDE USING btree ... WHERE ...`) is heavier than this app needs.

Pure deletions/shift-downs are safe today because the position-ordered association saves lower rows first — drag reorder introduces upward moves.

### 1. Server — sibling-aware validations
Persisted-duplicate checks exclude IDs of in-memory siblings (final values are covered by the existing `unsaved_*` in-memory checks):
- `ExercisePositionValidation`: `persisted_segment_position_exists?`, `persisted_segment_part_position_exists?`; replace the DB-only `validates :position, uniqueness:` for top-level exercises with a sibling-aware custom check (also adds the currently missing in-memory top-level vs top-level check).
- `Segment`: replace `uniqueness: { scope: :workout_id }`; make `persisted_top_level_exercise_position_exists?` sibling-aware.
- Single-record saves outside a workout graph (console, RailsAdmin) validate exactly as before.

### 2. Server — two-pass save to clear the unique indexes
`before_save` on `Workout`: bulk-negate (`update_all('position = -position')`, one update per scope) the positions of persisted children with a **pending position change**; autosave then writes the final positive values inside the same transaction. Positions are always positive, so negated values can't collide; every negated row is dirty, so autosave always rewrites it. No migration; indexes unchanged; rollback is a code revert.

⚠️ Segment children submitted via `segments_attributes` live on `segment.exercises`, not `workout.exercises` — collect dirty records from the graph nested attributes actually populated.

### 3. JS — `sortable_list_controller.js`
- `yarn add sortablejs`; new Stimulus controller (register in `controllers/index.js`) wrapping `Sortable.create` with `forceFallback: true` (touch support + the only drag mode Selenium can drive), `animation: 150`, per-level handle values (`.part-handle` top-level, `.exercise-handle` in segments) so grabbing a child handle can't drag the whole segment card.
- `renumber()`: walk `:scope > .nested-fields:not([hidden])`, write index+1 into each row's **own** position input (resolve via `input.closest('.nested-fields') === row` — segment cards contain child position inputs).
- `moveUp`/`moveDown` (insertBefore + renumber): keyboard-accessible fallback and the deterministic system-test path (SortableJS has no built-in keyboard support).
- Wire bubbling `nested-form:add`/`nested-form:remove` to `renumber` (idempotent).
- Cross-list dragging (exercise out of a segment / between segments) **out of scope**: separate Sortable instances, no shared `group`.

### 4. JS — simplify `nested_form_controller.js`
Remove `positionExercisesValue` / `positionSegmentsValue` / `assignPosition()` / `nextPosition()` — renumbering owns positions (contiguous after removals; gaps were an implementation artifact). Keep add/remove/dispatch (the metrics nested form depends on them).

### 5. Views
- `f.input :position` → `f.hidden_field :position` in `_exercise_fields` and `_segment_fields`.
- Grip handles (`faGripVertical` registered in `application.js`, FontAwesome `dom.watch` pattern) + small `aria-label`ed Move up/Move down buttons.
- Handle class passed as a partial local (`_exercise_fields` renders in both contexts).
- Minimal `.sortable-ghost` / `.sortable-chosen` styles in `workouts.scss`.

### 6. Tests
- Model: persisted position swaps via nested attributes succeed — top-level pair, within-segment pair, exercise↔segment swap, reorder combined with `_destroy` and a new appended part.
- System: rewrite `test/system/workout_segment_positions_test.rb` for hidden inputs (`all('input[name$="[position]"]', visible: :hidden)`); removal now renumbers contiguously (the "assigns positions past removed workout parts" expectation changes from gapped `1 3 4` to contiguous); reorder via Move buttons asserting hidden values and persisted order after submit; **one** `drag_to` smoke test only (flake-prone — drop it if CI flakes, keep button-driven coverage).
- Update `test/system/workout_workflows_test.rb` (~line 26) which asserts a visible Position field.

### 7. Docs & rollout
- Short `cf/docs/decisions.md` entry: positions become system-managed, contiguous-on-save, drag/move-button-driven UI.
- No migration or backfill; existing positions stay valid, saves close gaps opportunistically.
- Sequence within the PR: server steps 1–2 first with model tests (independently safe to land), then JS/views, then system tests.
- Keep `Segment#assign_position` as a backstop for console/seed creation.

## Risks

1. Unique-index swap is the core difficulty — steps 1–2 must land before the UI.
2. Capybara native drag is unreliable — rely on Move buttons for coverage.
3. Duplicate in-memory objects between `workout.exercises` and `segment.exercises` during nested assignment — worth an explicit model test.
4. Mobile: handles inside scrollable cards deserve a manual check; Move buttons double as the mobile-friendly path.

Yanchek99:
## Summary
- Replace manual workout position text fields with hidden position inputs driven by SortableJS drag ordering.
- Add a sortable-list Stimulus controller for top-level workout parts and segment-nested exercise lists.
- Add drag handles for exercise and segment rows/cards using the existing Font Awesome setup.
- Normalize visible sibling positions after drag/add/delete and preserve existing nested-form add/delete behavior.
- Reserve submitted persisted positions during workout updates so existing exercises/segments can swap positions without uniqueness/index collisions.

Closes #1638

## Validation
- yarn build
- yarn build:css (passes with existing Sass/Bootstrap deprecation warnings)
- rvm 3.4.9@wod-tracker do bundle exec rails test
- rvm 3.4.9@wod-tracker do bundle exec rails test test/system/workout_segment_positions_test.rb test/system/workout_workflows_test.rb test/system/exercise_card_editing_test.rb test/system/workout_drag_ordering_test.rb
- rvm 3.4.9@wod-tracker do bundle exec rubocop --parallel
- git diff --check


Yanchek99 (commented):
## Code Review

Solid, well-tested change — the reservation design is the right approach. Notes below; no correctness blockers.

### Correctness — the reservation logic

The core trick (`reserve_submitted_positions!` → set each submitted row to `position: -index` via `update_columns`, then run the real `update` inside one transaction with rollback) is correct, and reserving each scope separately correctly exploits that the three unique position indexes are disjoint. The swap regression is well covered by `workout_position_updates_test.rb` and the tabata system test.

A few invariants worth confirming:

- **Reservation runs on every update, even when nothing was reordered.** `update` now issues one `UPDATE` per exercise + per segment plus a `reload` on *every* save — renaming a 5-exercise workout rewrites all 5 rows. Functionally fine, just extra write churn; only worth a guard if workouts get large.
- **A submitted row missing its `position` is left negative until rollback.** If JS ever fails to populate a hidden `position`, `reserve_records` sets it to `-index`, the update doesn't overwrite it, and `numericality: greater_than: 0` fails the whole save → rollback. Safe (nothing persists), but the failure mode is now "entire update fails." The reservation assumes every submitted persisted row carries a position.
- **Partial submission.** The logic is correct only if the form submits *all* existing siblings (so they all get reserved). True for the current full-form render — just noting the dependency.

### Accessibility regression (worth a decision)

The old numeric Position inputs let keyboard/screen-reader users set ordering by typing. SortableJS drag is pointer/touch-only, so reordering is now **mouse/touch-only** — there's no keyboard path. The handles are nicely labeled, but the reorder action itself isn't keyboard-operable. If a11y matters here, consider arrow-key move on the handle or a focusable fallback.

### Frontend nits

- **`clearDragState` is fighting the library.** The triple `setTimeout([0, 50, 200])` to strip SortableJS leftover classes is timing-dependent, and the document-wide `querySelectorAll` means a drag in one list clears ghost/chosen classes in *every* sortable list on the page. It works, but a comment explaining *why* the delays are needed would help, and it'll be fragile across SortableJS bumps.
- **`segment-card` doesn't coordinate exclusive-open** the way `exercise-card` does (the `exercise-card:opening` event), so a segment and an exercise can be open simultaneously. Probably intentional — noting the asymmetry.
- **`segment-card#save` validity check spans nested exercise inputs.** `editorControls` includes inputs inside collapsed nested exercise cards; `reportValidity()` on a field inside a `hidden` container shows no visible bubble, so collapsing could silently no-op. Edge case (nested exercises self-validate on their own "Done").

### Tests

- Good coverage: reorder-on-create, mixed parts, segment exercises, edit-existing, the tabata cross-type swap, and handle visibility.
- The reorder helper uses `execute_script` to move DOM nodes and call `refresh()` directly — it exercises renumbering + server reservation but **not** the actual SortableJS drag (handle binding, `filter`, ghost classes). Reasonable tradeoff, just noting the library wiring itself is untested.
- The `turbo_conventions_test` switch to `[data-controller~="nested-form"]` is the correct fix for the now-multi-controller attribute.

### Bundled-in change

Removing `assert_text 'Workout was successfully created.'` from `workout_workflows_test.rb` is a reasonable flake fix (the commented Turbo flash race), but it's unrelated to drag ordering and slightly weakens that assertion. Fine to keep — just flagging it rides along.

### Verdict

Approve with minor follow-ups. The two I'd actually consider acting on: the **keyboard-reorder accessibility gap** and a **comment documenting the `clearDragState` timeout hack**. Everything else is optional polish.

Yanchek99 (commented):
Reviewed the drag-ordering change. The position-reservation design is sound and the swap regression is well covered. Left inline notes — no correctness blockers. The two I'd actually act on are the keyboard-reorder accessibility gap and a comment on the `clearDragState` timing hack.

app/controllers/workouts_controller.rb
```diff
@@ -37,8 +37,18 @@ def create
   # PATCH/PUT /workouts/1
   # PATCH/PUT /workouts/1.json
   def update
+    attributes = workout_params
+    updated = false
+
+    @workout.transaction do
+      @workout.reserve_submitted_positions!(attributes)
```
Yanchek99, line 44: Reservation runs on **every** update, even when nothing was reordered — one `UPDATE` per exercise + per segment plus a `reload` on every save (renaming a 5-exercise workout rewrites all 5 rows). Functionally fine, just extra write churn. Only worth guarding (e.g. skip when no submitted position differs from persisted) if workouts get large.

app/models/concerns/workout_position_reservation.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def existing_segment_exercises(segment_id, segment_attributes)
+    exercise_ids = nested_attribute_ids(segment_attributes[:exercises_attributes])
+    Exercise.unscoped.where(segment_id:, id: exercise_ids)
+  end
+
+  def reserve_records(records)
+    records.each.with_index(1) do |record, index|
+      # Reserve submitted rows inside the update transaction so uniqueness indexes allow position swaps.
+      record.update_columns(position: -index) # rubocop:disable Rails/SkipsModelValidations
```
Yanchek99, line 38: A submitted persisted row that arrives **without** a position is left at `-index` here, and since the subsequent `update` never overwrites it, `numericality: greater_than: 0` fails the whole save → rollback. Safe (nothing persists), but the failure mode becomes "entire update fails" rather than a localized error. This reserve logic depends on the invariant that every submitted persisted sibling carries a position — true for the current full-form render, worth a comment so it isn't broken later.

app/javascript/controllers/sortable_list_controller.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  refresh() {
+    this.items.forEach((item, index) => {
+      const positionInput = item.querySelector('input[name$="[position]"]')
+      if (positionInput) positionInput.value = index + 1
+    })
+  }
+
+  clearDragState(item) {
```
Yanchek99, line 60: This is fighting SortableJS's own cleanup: the triple `setTimeout([0, 50, 200])` is timing-dependent, and `dragClassElements` does a document-wide `querySelectorAll`, so a drag in one list strips ghost/chosen classes from **every** sortable list on the page. It works, but it'll be fragile across SortableJS bumps. At minimum add a comment explaining why the delays are needed; ideally scope the cleanup to this controller's container.

app/javascript/controllers/segment_card_controller.js
```diff
@@ -0,0 +1,109 @@
...
+      const element = document.createElement("span")
+      element.className = "segment-summary__detail"
+      element.textContent = summary
+      return element
+    }))
+    this.summaryDetailsTarget.hidden = this.exerciseSummaryTexts.length === 0
+  }
+
+  get editorControls() {
+    return Array.from(this.editorTarget.querySelectorAll("input, select, textarea"))
```
Yanchek99: `editorControls` selects inputs inside collapsed nested exercise cards too. `save()` calls `reportValidity()` on the first invalid one, but a field inside a `hidden` container shows no visible validation bubble — so collapsing the segment could silently no-op with no feedback. Edge case (nested exercises self-validate on their own "Done"), but worth scoping this to the segment's own controls.

app/views/workouts/_segment_fields.html.slim
```diff
@@ -1,25 +1,43 @@
...
-          = f.simple_fields_for :exercises, Exercise.new, child_index: 'CHILD_EXERCISE' do |exercise_form|
-            = render 'exercise_fields', f: exercise_form
-        .links.d-grid.gap-2
-          = link_to 'Add Exercise', '#', class: 'btn btn-sm btn-outline-primary', data: { action: 'click->nested-form#add' }
-    .row
-      .col.d-grid
-        = link_to 'Delete Segment', '#', class: 'btn btn-sm btn-danger mt-2', data: { action: 'click->nested-form#remove' }
+  = f.hidden_field :position
+  .segment-summary.card-body data-segment-card-target="summary"
+    button.workout-sortable-handle.segment-summary__handle type="button" aria-label="Drag segment"
```
Yanchek99: Accessibility regression: the old numeric Position inputs let keyboard/screen-reader users set ordering by typing. SortableJS drag is pointer/touch-only, so reordering is now mouse/touch-only — the handle is labeled but the reorder action itself isn't keyboard-operable. Consider arrow-key move on the handle (or a focusable fallback) if a11y matters here. Same applies to the exercise handle.

test/system/workout_drag_ordering_test.rb
```diff
@@ -0,0 +1,187 @@
...
+      done_button = find_button('Done')
+      scroll_to done_button, align: :center
+      done_button.click
+    end
+  end
+
+  def reorder_sortable_list(root, from:, to:)
+    root = find(root) if root.is_a?(String)
+
+    page.execute_script(<<~JS, root, from, to)
```
Yanchek99, line 51: This drives the reorder via `execute_script` (move DOM node + call `refresh()`), which exercises the renumbering and server-side reservation but **not** the actual SortableJS drag — handle binding, the `filter` selector, and ghost-class behavior go untested. Reasonable tradeoff vs. flaky drag sims, just noting the library wiring has no automated coverage.

test/system/workout_workflows_test.rb
```diff
@@ -33,8 +33,10 @@ class WorkoutWorkflowsTest < ApplicationSystemTestCase
 
     click_on 'Create Workout'
 
+    # Landing on the workout show page with its name and exercise proves the
```
Yanchek99, line 36: Dropping the `'Workout was successfully created.'` assertion is a reasonable flake fix (the commented Turbo flash race), but it's unrelated to drag ordering and slightly weakens this test. Fine to keep — just flagging it rides along in this PR.

app/controllers/workouts_controller.rb
Yanchek99: Leaving this as-is for now. The extra reservation writes are real, but the current form sizes are small and the unconditional path keeps the swap/reservation logic simple. I would add a guard only if this shows up as actual write churn on larger workouts.

app/models/concerns/workout_position_reservation.rb
Yanchek99: Fixed in `95d5ebf`. Added a comment to `WorkoutPositionReservation#reserve_records` documenting the invariant that the full workout form submits every persisted sibling with a hidden positive `position`, and that validation rollback protects us if that invariant is broken.

app/javascript/controllers/sortable_list_controller.js
Yanchek99: Fixed in `95d5ebf`. `clearDragState` now has an explanatory comment for the delayed cleanup, and `dragClassElements` is scoped to this controller's container instead of querying the whole document.

app/javascript/controllers/segment_card_controller.js
Yanchek99: Fixed in `95d5ebf`. Segment save/collapse validation now targets only the segment's own fields (`name`, `rounds`, `time_seconds`, `interval_scheme`, `rest_seconds`, `notes`) instead of querying every input under the editor, so collapsed nested exercise inputs no longer participate.

app/views/workouts/_segment_fields.html.slim
Yanchek99: Fixed in `95d5ebf`. Added keyboard reordering on both exercise and segment handles: focused handles respond to ArrowUp/ArrowDown, move the current card within its sortable list, refresh hidden positions, and keep focus on the moved handle. Also added a focused system test for the keyboard path.

test/system/workout_drag_ordering_test.rb
Yanchek99: Leaving this tradeoff as-is for now. I agree the existing drag-ordering tests primarily cover DOM order normalization plus server-side position persistence, not browser-level SortableJS drag simulation. I added keyboard-handle coverage in `95d5ebf`, but I did not add pointer drag simulation because those system tests tend to be brittle across drivers/browser timing.

test/system/workout_workflows_test.rb
Yanchek99: Leaving this as-is. The flash assertion removal is unrelated to drag ordering, but it was intentionally kept because that assertion was racing Turbo flash rendering. The workflow still asserts the create path and persisted workout behavior.

Issue #1673, opened 2026-06-18
Idempotency: source columns on workouts + CrossFit.com program


Yanchek99:
Make WOD imports re-runnable without duplicates (part of the WOD-scraper epic).

## Scope
- Migration adding `source` (string) and `source_ref` (string, the `YYMMDD` slug) to `workouts`, with a unique composite index; keep `db/schema.rb` in sync.
- Seed/find a dedicated `Program` (e.g. "CrossFit.com Main Site") owned by the admin user; scraped workouts are scheduled onto it via `Schedule.posted_at` (the WOD date).
- Importer upserts via `Workout.find_or_initialize_by(source: "crossfit.com", source_ref: slug)` and `program.schedules.find_or_initialize_by(workout:).update(posted_at:)`.

Mirror the existing build pattern in `db/seeds.rb` (e.g. Fight Gone Bad L179, Barbara L225) and its `schedules.find_or_initialize_by(...).update(posted_at:)` usage.

## Acceptance
- Running an import twice for the same date yields exactly one workout and one schedule.

Yanchek99:
Closing: superseded by mechanisms that landed after this was filed.

The goal here was idempotent WOD imports (re-run → one workout, one schedule). That's now covered without dedicated `source`/`source_ref` columns on `workouts`:

- **Workout reuse** → `content_key` fingerprint (`WorkoutFingerprint`). Re-importing identical content finds/absorbs the existing workout instead of creating a duplicate.
- **Source** → the dedicated "CrossFit.com Main Site" `Program` *is* the origin.
- **Date / import cursor** → `Schedule.posted_at` already stores the WOD date (the same value the `YYMMDD` slug encodes). "Have I scraped this date?" is just `program.schedules.exists?(posted_at: date)`.

So the importer upserts on `content_key` for the workout and `(program_id, posted_at)` for the schedule — both already exist and `merge_into!` already dedupes schedules on `(program_id, posted_at)`.

A unique `source_ref` on `workouts` would also have actively fought the desired behavior: the same workout (e.g. Fran) appears on multiple dates and should reuse one row, but `source_ref` being the date slug would force a separate row per date.

Upstream URL isn't needed as a key; if wanted for humans it can live in notes.

PR #1682, opened 2026-06-18
Idempotency: content-fingerprint workout identity + merge-on-duplicate

Yanchek99:
Toward #1673 (WOD-scraper epic): make imports re-runnable and workouts deduplicated by content. Identity is a content fingerprint; duplicates are merged rather than rejected.

## Workout identity = content, not date

A workout's identity is its prescribed content, independent of when it's scheduled. A benchmark repeated on many dates (Fran, Cindy, Murph…) is **one `Workout` on many `Schedule`s**.

- **`workouts.content_key`** — a SHA-256 fingerprint of the scoring scheme and ordered parts (exercises/segments + prescriptions), with a **global unique index**. Computed in `before_save`; refreshed `after_commit` when child exercises/segments change directly. Free-text notes are excluded (see #1684 for the canonical-load gap).
- The date lives only on `Schedule#posted_at`.

## Duplicates merge instead of erroring

Editing or creating a workout into the same content as an existing one yields a single workout:

- On save, a workout whose content matches an existing one keeps a **nil** `content_key` (so it never hits the unique index).
- `WorkoutsController#create`/`#update` then call **`Workout#absorb_duplicate!`**, which folds the duplicate's **schedules** (deduped on `program` + `posted_at`) and **logs** into the canonical workout, deletes the duplicate, and **redirects to the survivor**.

## CrossFit.com program seed

Seeds a single **`Crossfit.com`** `Program` owned by the admin user — the curated benchmark library and the destination for scraped WODs.

## Docs (source-grounded in L1/L2)

- `cf/docs/terminology.md` + `programming.md`: benchmark workouts as a repeatable fitness *test*; repeat-workout modeling rule.
- `cf/docs/decisions.md`: repeat/benchmark = one Workout on many Schedules, the content-fingerprint + merge mechanism, and the deferred canonical-load identity (#1684).

## Follow-ups (out of scope)

- **No importer yet** — this PR delivers the fingerprint + merge capability. The `find_or_initialize_by(content_key:)` upsert lands with the importer issue.
- **Load canonicalization (#1684)** — until lb/kg/pood normalize, the fingerprint uses stored load values (safe for the lb-consistent CrossFit.com source).

## Verification

- Migration runs + backfills; 35 seed workouts get distinct keys. Schema in sync.
- Model tests (`workout_fingerprint_test.rb`): fingerprint stability, nil-on-duplicate, child-change refresh, marked-for-destruction exclusion, and `absorb_duplicate!` merging schedules (deduped) + logs and deleting the duplicate.
- Request test (`workout_merge_test.rb`): editing a workout into an existing one redirects to the canonical and removes the duplicate.
- Model + system suites green locally (incl. drag-ordering and workflows). RuboCop clean.
- Also includes an AGENTS.md Ruby version fix (3.4.9 → 4.0.5).


app/models/concerns/workout_fingerprint.rb
```diff
@@ -0,0 +1,74 @@
+module WorkoutFingerprint
+  extend ActiveSupport::Concern
+
+  included do
+    before_save :assign_content_key
```
Yanchek99, line 5: This callback only runs when the parent `Workout` is saved, but the key is derived from child `Exercise` and `Segment` rows. Creating an empty workout and then adding/updating/deleting parts directly leaves `content_key` stale; nested destroys can also compute from records marked for deletion before autosave removes them. The uniqueness guarantee needs the key to be refreshed when the content-bearing child records change.

app/models/concerns/workout_fingerprint.rb
```diff
@@ -0,0 +1,74 @@
...
+        rounds: segment.rounds,
+        time_seconds: segment.time_seconds,
+        interval_scheme: segment.interval_scheme,
+        rest_seconds: segment.rest_seconds,
+        exercises: segment.exercises.map { |exercise| canonical_exercise(exercise) }
+      }
+    }
+  end
+
+  def canonical_exercise(exercise)
```
Yanchek99, line 106: The fingerprint omits `Exercise#notes`, but notes are already used as prescription data in seeds, for example Linda's bodyweight loads. Two otherwise identical workouts with different prescription notes will collide on the same `content_key`, and the unique index will reject a legitimate distinct workout.

Yanchek99 (commented):
I found two correctness issues that should be addressed before merging.

app/models/concerns/workout_fingerprint.rb
Yanchek99: Fixed in b42c1bb. Added a `RefreshesWorkoutContentKey` concern on `Exercise` and `Segment` that recomputes the parent's key `after_save`/`after_destroy`, so direct child create/update/delete no longer leaves `content_key` stale. `content_fingerprint` now also excludes `marked_for_destruction?` records (nested destroys fingerprint the resulting content, not the rows about to be removed), and `refresh_content_key!` resets the cached associations so it always reads the current persisted parts. Covered by new tests: direct exercise change/destroy, adding a part to an empty workout, and the marked-for-destruction exclusion.

app/models/concerns/workout_fingerprint.rb
Yanchek99: Fixed in b42c1bb. `Exercise#notes` (and `Segment#notes`) are now part of the fingerprint, so Linda-style prescription notes no longer collide. New test `prescription notes change the fingerprint` covers it. I left workout-level rich-text `notes` out deliberately — those are descriptive prose (e.g. the Fight Gone Bad/Hope write-ups), and including them would split legitimately-identical prescriptions; the existing seeds are already structurally distinct, so this doesn't reintroduce a collision. Happy to fold workout notes in too if you'd prefer maximum strictness.

Yanchek99: Update: reverted this in 200a114 — notes are now kept **out** of the fingerprint after all. Free-text isn't a stable identity field, and the real case here (bodyweight/percentage loads stuffed into notes, like Linda's `1 1/2 body weight`) is the strength-percentage/canonical-load gap tracked in #1684 — it should become a proper load field that's naturally part of the fingerprint, not hashed prose. Locked the decision with a test asserting notes don't change the fingerprint. Tradeoff acknowledged: until #1684, two workouts identical except for a bodyweight-multiplier note would collide; none exist in the current seeds.

db/seeds.rb
```diff
@@ -161,7 +161,12 @@
 Movement.find_or_create_by(name: 'Zercher Squat')
 
 # Programs
-cfj = Program.find_or_create_by(name: 'Crossfit Journal')
+# Crossfit.com is the curated benchmark library seeded below and the destination
+# for scraped main-site WODs, owned by the admin user.
```
Yanchek99: remove this comment
```suggestion
```

Issue #1684, opened 2026-06-18
Model prescribed load as a canonical magnitude (lb/kg/pood are display conventions)


Yanchek99:
Part of the WOD-scraper epic. Prerequisite for content-based workout deduplication (see #1673 and the `cf/docs/decisions.md` repeat/benchmark + load-identity entries).

## Problem

CrossFit expresses a single prescribed load in whichever unit suits the audience — **pounds, kilograms, or pood** — and these are *display conventions for the same prescription*, not different prescriptions. The current model stores load as a value plus a `load_unit` enum on `exercises`, which folds the display unit into the load's **identity**. The prescribed *magnitude* is the identity; lb/kg/pood are presentations of it.

This blocks reliable dedup: a content fingerprint only guarantees uniqueness up to its canonical form, so the same prescription stored as both `95 lb` and `43 kg` would produce two `Workout` rows.

## Why it can't be fixed by unit conversion alone

The equivalences are **rounded conventions tied to standard implements**, not exact arithmetic:

| pood | implement | CrossFit lb label | exact lb |
|------|-----------|-------------------|----------|
| 1 | 16 kg | 35 lb | 35.3 |
| 1.5 | 24 kg | 53 lb | 52.9 |
| 2 | 32 kg | 70 lb | 72.2 |

Barbell Rx similarly publishes pairs like `95 lb / 43 kg` (43 kg ≈ 94.8 lb). Canonicalizing by physics (everything → grams) would *split* `95 lb` from `43 kg`, which is backwards. Canonicalization must follow CrossFit's **published standard-equivalence table**.

## Direction (decide in this issue)

- **Option A — canonical magnitude + equivalence table.** Store load as one canonical magnitude; normalize lb/kg/pood inputs through a source-confirmed CrossFit standards table; display unit becomes a presentation/user-preference concern. Lighter change (migration to drop unit from identity + a normalization helper + importer wiring).
- **Option B — `StandardLoad` concept.** First-class standard loads carrying lb/kg/pood labels per implement, referenced by exercises for Rx loads; arbitrary/user loads keep a raw magnitude. Most faithful, larger model change.

## Constraints

- The exact equivalence table (pood values, lb/kg Rx pairs, standard kettlebell implements) **must be source-confirmed from the L1/L2 guides and CrossFit prescriptions** before encoding, per `cf/docs/OVERVIEW.md`. The values above are illustrative and not yet source-verified.
- Existing user-created workouts must not be silently rewritten.
[...]

Issue #1631, opened 2026-06-05
Epic: Data-preserving schema redesign for scaling foundation


Yanchek99:
## Summary

Track the full data-preserving schema redesign for WOD Tracker. The goal is to keep PostgreSQL, replace polymorphic `metrics` with structured domain columns, preserve existing data through additive migrations/backfills, support hundreds of programs posting daily workouts, support lazy user-specific workout instances, and prepare for future scaling recommendations from movement-level logged history.

## Implementation Issues

- [x] #1622
- [x] #1623
- [x] #1624
- [x] #1625
- [x] #1626
- [x] #1627
- [ ] #1628
- [x] #1629
- [ ] #1630

## Recommended Order

1. Add direct workout and log score fields (#1622).
2. Clean up segment modeling (#1623).
3. Move exercise prescriptions onto exercises (#1624).
4. Move movement log performance onto movement logs (#1625).
5. Remove legacy metrics storage (#1626).
6. Evolve schedules into program workouts (#1627).
7. Add lazy user workout instances (#1628).
8. Add movement taxonomy and substitutions (#1629).
9. Build initial rule-based scaling foundation (#1630).

## Data Preservation Rules

- Each phase should be additive before destructive.
- Backfills must copy existing data from old fields/tables into new fields/tables.
- Reads can temporarily prefer new columns and fall back to legacy metrics during migration.
- Writes should switch to new columns only after backfill.
- Legacy metric storage should be removed only after tests and parity checks pass.
- Any unmapped metric should be reported during migration instead of silently dropped.

## Acceptance Criteria

- All child issues are linked from this epic.
- Overall migration order is clear.
- Destructive cleanup only happens after parity is verified.
- Existing seeded and user-created data is preserved.
- The resulting schema is structured enough to support future movement-level scaling recommendations.

## Assumptions

- PostgreSQL remains the source-of-truth database.
- The first goal is structured data quality, not immediate ML.
- Constantly varied workouts mean recommendations should generalize from movements, movement families, loads, volume, time domains, and logged performance.
- Personalized workout copies should be created lazily, only after user engagement.

Issue #1630, opened 2026-06-05
Build initial rule-based scaling foundation


Yanchek99:
## Summary

Build the first inspectable scaling recommendation layer from movement-log history before introducing ML.

## Work

- Use movement-log history to generate initial recommendations.
- Store scaling rationale on workout instance exercises.
- Start with Grace-style load scaling as the first proving case.
- Keep recommendations auditable by comparing Rx, generated prescription, and actual performance.

## Acceptance Criteria

- The app can recommend a user-specific load from prior movement history.
- Recommendation is inspectable and does not mutate Rx.
- Data shape is ready for later ML training.

## Notes

This should stay rule-based initially. The first goal is to validate that the structured schema can support useful scaling decisions before adding ML complexity.

Yanchek99:
## Two design decisions to settle before this starts (surfaced reviewing #1688 / #1629)

The taxonomy PR (#1688) lands the data layer this issue builds on. Two choices are cheap to make now and expensive to retrofit once a scaler consumes them:

### 1. History-fallback grain — don't fall back on `family`

This issue leads with Grace-style load scaling. When an athlete's exact-movement history for a movement (e.g. Clean and Jerk) is thin, the scaler needs a fallback. #1688 provides `movement_family_history_for`, but `family` has only 4 values (mono/gymnastics/weightlifting/rest), so a C&J fallback returns the athlete's *entire weightlifting history* — deadlift, bench, snatch, back squat — which isn't a defensible basis for a C&J load.

Proposal: fall back on **function + equipment** (e.g. barbell hinge/pull-under movements), not modality. The data already exists in `movement_function_assignments` + `equipment`; it just needs a query helper. Suggest defining the fallback grain here so #1688's family helper isn't mistaken for the scaling fallback.

### 2. Substitutions — derive vs. hand-curate

#1688 adds a directed `movement_substitutions` table but seeds it with nothing (constantly-varied WODs make hand-authoring every easier/harder/lateral pair unscalable). Recommend the scaler **derive** candidate substitutes from shared primary function + family + equipment + adjacent skill level, and reserve the manual table for curated overrides/exclusions. Deriving is also the more ML-ready shape (it's a similarity function, not a fixed edge list). Worth deciding before this issue treats the manual table as the source of truth.

### Related label-quality note

[...]

Issue #1628, opened 2026-06-05
Add lazy user workout instances


Yanchek99:
## Summary

Add lazily generated user-specific workout instances so personalized scaling can exist without mutating canonical Rx workouts or pre-generating copies for every subscriber.

## Work

- Add `workout_instances`.
- Add `workout_instance_exercises`.
- Generate only when a user opens, starts, edits, accepts, or logs a workout.
- Link logs to workout instances.
- Copy canonical prescription fields into instance exercises when generated.

## Acceptance Criteria

- No subscriber-wide fan-out generation.
- Canonical Rx, generated recommendation, and actual performance can be compared.
- Canonical workouts are never mutated for personalization.
- Logs can identify the generated workout instance they came from.

## Notes

This issue should come after canonical workout and exercise data is structured enough to copy cleanly.

PR #1688, opened 2026-06-21
[codex] Add movement taxonomy and substitutions

### LINKED ISSUE #1629: Add movement taxonomy and substitutions
## Summary

Add structured movement intelligence so future scaling can generalize across constantly varied workouts and recommend substitutions, not only numeric changes.

## Work

- Add movement family, pattern, equipment, skill, and load-bearing metadata.
- Add structured substitutions with easier/harder/lateral direction.
- Support exact movement history queries and movement-family fallback queries.

## Acceptance Criteria

- User history can be queried by exact movement and movement family.
- Future scaling can recommend substitutions, not only numeric changes.
- Movement metadata remains structured and queryable.

## Notes

This is part of the ML/scaling foundation. It should be informed by CrossFit domain docs and updated in `cf/docs/` as durable terminology/modeling decisions are clarified.

Yanchek99:
## Summary

- Add structured movement taxonomy fields for family, pattern, equipment, skill level, and load-bearing metadata.
- Add directed movement substitutions with easier, harder, and lateral directions.
- Add exact movement and movement-family history query helpers for future scaling work.
- Document the movement taxonomy modeling decision in `cf/docs/`.

Closes #1629.

## Validation

- `rvm 4.0.5@wod-tracker do bundle exec rails test test/models`
- `rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel`

Note: the repository agent guide still mentions Ruby 3.4.9, but the current Gemfile requires Ruby 4.0.5, so validation used `rvm 4.0.5@wod-tracker`.


Yanchek99:
@claude review

Yanchek99 (commented):
**🤖 Code review by Claude (Claude Code), posted on Matt's behalf.**

Solid, well-tested addition. The taxonomy/substitution modeling is clean, the migration and `db/schema.rb` are in sync, FKs and the directed join table are correct, and the decision doc + standards updates capture the rationale well. Tests cover the new scopes, the self-substitution guard, and the enum accessors. I have no blocking concerns — the inline notes below are minor/optional.

**Things I verified as correct (not issues):**
- `MovementLog.for_movement_family` passes an enum *label* string (`movement.family` → `"gymnastics"`) into `where(movements: { family: })`. Rails 8 type-casts nested association conditions via the joined model's attribute types, so the enum label maps to the integer correctly — confirmed by the passing family-history tests.
- Both `movement_substitutions` and `inverse_movement_substitutions` are `dependent: :destroy`, so destroying a movement won't orphan substitution rows on either side of the FK.
- The DB unique index `[movement_id, substitute_movement_id, direction]` matches the model-level uniqueness scope.

**Minor / optional (inline):** index selectivity on low-cardinality columns, contradictory substitution directions, and `family_movements` including the movement itself.

db/migrate/20260620130000_add_movement_taxonomy_and_substitutions.rb
```diff
@@ -0,0 +1,28 @@
...
+    add_column :movements, :pattern, :integer
+    add_column :movements, :equipment, :integer
+    add_column :movements, :skill_level, :integer
+    add_column :movements, :load_bearing, :boolean, null: false, default: false
+
+    add_index :movements, :family
+    add_index :movements, :pattern
+    add_index :movements, :equipment
+    add_index :movements, :skill_level
+    add_index :movements, :load_bearing
```
Yanchek99: 🤖 Claude: Indexing `load_bearing` (a 2-value boolean) is unlikely to earn its keep — Postgres will typically seq-scan rather than use a single-column index on a column with so few distinct values, so this mostly adds write/storage overhead. The enum columns (`family`, `pattern`, `equipment`, `skill_level`) are also low-cardinality; the `family` index is the one with a clear consumer (the `for_movement_family` join filter), but the others have no querying code yet in this PR. Consider dropping `load_bearing` (and possibly `skill_level`/`equipment`/`pattern`) until there's a query that needs them. Non-blocking.

app/models/movement_substitution.rb
```diff
@@ -0,0 +1,24 @@
...
+    lateral: 2
+  }.freeze
+
+  belongs_to :movement
+  belongs_to :substitute_movement, class_name: 'Movement'
+
+  enum :direction, DIRECTIONS, prefix: true
+
+  validates :direction, presence: true
+  validates :substitute_movement_id, uniqueness: { scope: %i[movement_id direction] }
```
Yanchek99: 🤖 Claude: Because `direction` is part of the uniqueness scope, the same ordered pair can be stored as both `easier` and `harder` (e.g. pull-up → push-up recorded as easier *and* harder), which is contradictory. If a given (movement, substitute) pair should only ever have one direction, scope the uniqueness to just `movement_id` instead and drop `direction` from both this validation and the DB index. If multiple directions per pair is intentional, ignore this. Non-blocking.

app/models/movement.rb
```diff
@@ -12,4 +71,10 @@ def self.search_by_name(name)
     end
     where(query)
   end
+
+  def family_movements
```
Yanchek99: 🤖 Claude: `family_movements` returns `where(family:)`, which includes the movement itself. For the family-fallback history use case that's fine (you'd want the movement's own logs included), but the name reads like "other movements in the same family." Worth a one-line comment clarifying that self is intentionally included, or `.where.not(id:)` if it isn't. Non-blocking.

db/migrate/20260620130000_add_movement_taxonomy_and_substitutions.rb
Yanchek99: Addressed in cc80d15: removed the low-selectivity single-column indexes for pattern, equipment, skill_level, and load_bearing. Kept the family index because it supports the family-history query path.

app/models/movement_substitution.rb
Yanchek99: Addressed in cc80d15: changed both the model validation and DB unique index so each ordered movement/substitute pair can have only one direction.

app/models/movement.rb
Yanchek99: Addressed in cc80d15: added a short comment clarifying that family_movements intentionally includes self so family fallback history includes exact movement history too.

db/migrate/20260620130000_add_movement_taxonomy_and_substitutions.rb
Yanchek99: Follow-up in d8412c7: restored pattern, equipment, and skill_level indexes. This issue is part of the broader ML/scaling foundation, so those taxonomy dimensions are expected to become query dimensions in follow-up work even though this PR only wires the first family query. The load_bearing boolean index remains removed because it is still low-selectivity.

Yanchek99:
Updated the seed data in d88cb00 so seeded movements now get taxonomy metadata through an idempotent helper. This also updates existing seeded movement rows instead of only filling taxonomy on first creation.

A few details based on the taxonomy feedback:
- No seed entries use mixed modality/family.
- No seed entries use equipment values for bodyweight/no equipment.
- No seed entries use mixed equipment; scalar equipment picks the main loaded implement where applicable.
- Loaded movements are seeded as weightlifting; unloaded gymnastics/monostructural movements remain in their respective families.

Verification run:
- rvm 4.0.5@wod-tracker do ruby -c db/seeds.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel db/seeds.rb app/models/movement.rb
- git diff --check

I also tried rvm 4.0.5@wod-tracker do bundle exec rails db:seed, but local PostgreSQL is still refusing connections on ::1:5432 and 127.0.0.1:5432, so Rails aborts before executing the seed file.

Yanchek99:
Updated the seed data per the cleanup feedback in 34063b9. The seed_movement helper is gone; each movement now uses Movement.find_or_initialize_by(...).tap directly, assigns taxonomy inline, and saves the record.

Verification run:
- rvm 4.0.5@wod-tracker do ruby -c db/seeds.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel db/seeds.rb
- git diff --check

Yanchek99:
Updated the movement substitution fixtures in c465d3e to use the requested representative pairs:

- backsqt_frontsqt: front_squat -> back_squat, direction 0
- run_row: run -> row, direction 2

I added the missing front_squat movement fixture and updated the model test assertions to match.

Verification run:
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_substitution_test.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel test/models/movement_substitution_test.rb
- git diff --check

Yanchek99:
Corrected AbMat Sit-up in 2bd4128: its seeded movement pattern is now hinge instead of rotation.

Verification run:
- rvm 4.0.5@wod-tracker do ruby -c db/seeds.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel db/seeds.rb
- git diff --check

Yanchek99:
Updated in aff02d0: removed `squat` as a movement pattern and renamed that enum value to `lower_body_press`. Squat-style movements in the seed data now use `pattern: :lower_body_press`, while upper-body press movements stay on `pattern: :press`.

Verification run:
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_substitution_test.rb
- rvm 4.0.5@wod-tracker do ruby -c db/seeds.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel app/models/movement.rb test/models/movement_test.rb test/models/movement_substitution_test.rb db/seeds.rb
- git diff --check

Yanchek99:
Updated in 998acae: movement patterns are now multi-valued instead of a single enum column.

What changed:
- Replaced `movements.pattern` with `movements.patterns` as an integer array with a GIN index.
- Removed the generic `mixed` pattern from movement taxonomy.
- Split force patterns into `lower_body_press`, `lower_body_pull`, `upper_body_press`, and `upper_body_pull`.
- Added model helpers/querying for component patterns, e.g. `pattern_lower_body_press?` and `Movement.with_pattern(:upper_body_press)`.
- Updated seeds so compound movements carry explicit component patterns. Examples: thrusters use lower-body + upper-body press; squat snatch/snatch use lower-body pull, upper-body pull, upper-body press, and lower-body press.
- Updated fixtures/tests and taxonomy docs to match.

Verification run:
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_substitution_test.rb
- rvm 4.0.5@wod-tracker do ruby -c db/seeds.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel app/models/movement.rb test/models/movement_test.rb test/models/movement_substitution_test.rb db/seeds.rb
- git diff --check

Yanchek99:
Updated in 5d95656 to align movement patterns with CrossFit movement functions from Programming Basics Part 1.

What changed:
- Replaced the lower/upper-body force labels with function terms: squat, hinge, vertical/horizontal push, vertical/horizontal pull, trunk flexion/extension/rotation, carry, locomotion, inversion, jump, and rest.
- Kept patterns multi-valued, so compound movements still carry each meaningful function.
- Updated seed assignments for clear horizontal pushes/pulls, trunk extension/rotation, and inversion + vertical push cases.
- Added the CrossFit source link to the movement taxonomy docs.

Source: https://www.crossfit.com/pro-coach/programming-basics-part-1

Verification run:
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_substitution_test.rb
- rvm 4.0.5@wod-tracker do ruby -c db/seeds.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel app/models/movement.rb test/models/movement_test.rb test/models/movement_substitution_test.rb db/seeds.rb
- git diff --check

Yanchek99:
Updated in cb4286b: renamed the movement taxonomy from patterns to functions across the model API, migration/schema, seed data, fixtures, tests, and docs. The taxonomy still uses the CrossFit function vocabulary and keeps functions multi-valued for compound movements like thrusters and snatches.

Verification:
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_substitution_test.rb
- rvm 4.0.5@wod-tracker do ruby -c db/seeds.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel app/models/movement.rb test/models/movement_test.rb test/models/movement_substitution_test.rb db/seeds.rb
- git diff --check

Yanchek99:
Cleaned this up in 7fb8ddd. I removed the inferred function values that were not in the CrossFit source: carry, locomotion, inversion, jump, rest, and trunk_rotation.

The Movement function taxonomy is now limited to the sourced list: squat, hinge, vertical_push, vertical_pull, horizontal_push, horizontal_pull, trunk_flexion, and trunk_extension. Seed movements that only had an inferred function now get functions: [] so reseeding clears unsupported values instead of preserving stale taxonomy data.

Verification:
- rvm 4.0.5@wod-tracker do ruby -c db/seeds.rb
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_substitution_test.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel app/models/movement.rb test/models/movement_test.rb test/models/movement_substitution_test.rb db/seeds.rb
- git diff --check

Yanchek99:
Updated in 6ffe8d2 to model movement functions with roles instead of a flat array.

What changed:
- Replaced the movements.functions array with movement_function_assignments.
- Each assignment stores movement_function plus role: primary, secondary, or tertiary.
- Added role-aware query support via with_function and with_function_role.
- Updated fixtures so thruster has primary squat and secondary vertical_push.
- Updated seed data so representative compound lifts like cleans, snatches, thrusters, and wall balls separate primary and secondary phases instead of treating all functions equally.
- Updated docs to avoid contribution weights and describe categorical roles.

I kept this on Movement rather than Exercise because in this app Exercise is the workout occurrence of a Movement; the reusable taxonomy belongs to the movement record.

Verification:
- rvm 4.0.5@wod-tracker do ruby -c db/seeds.rb
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_substitution_test.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel app/models/movement.rb app/models/movement_function_assignment.rb app/models/concerns/movement_function_assignable.rb test/models/movement_test.rb test/models/movement_substitution_test.rb db/seeds.rb
- git diff --check

Yanchek99:
Fixed in 1fffd30. I renamed the migration from 20260625120000 to 20260701120000 so it sorts after the current master migrations, which now include 20260626000000_add_ascending_ladder_columns and 20260630120000_add_team_size_to_workouts.

I also updated db/schema.rb to version 2026_07_01_120000 and carried forward master’s ladder/team schema columns so this PR no longer looks like it removes those master schema changes.

Verification:
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_substitution_test.rb
- git diff --check

Yanchek99:
## Review — movement taxonomy & substitutions

Reviewed against #1629's acceptance criteria (not the whole ML goal — per Epic #1631 this is the data-quality layer, with the scaler in #1630). Solid, well-tested foundation. Exact-movement history + structured metadata land cleanly and unblock the Grace load-scaling case in #1630. A few things worth addressing before #1630 builds on top.

### Scoped to this PR

- **Family fallback is too coarse to be useful (highest-value item).** `movement_family_history_for` keys only off `family`, which has 4 values. For the clean-and-jerk in Grace, the "family fallback" returns the athlete's *entire weightlifting history* (deadlift, bench, snatch, back squat…), which you'd never use to infer a C&J load. The data to do a finer fallback (function + equipment) already exists in `movement_function_assignments` but the helper ignores it. See the linked note on #1630 — I'd like to settle the fallback grain there before this is consumed.
- **Substitution graph is empty.** Seeds populate zero substitutions (only 2 in fixtures), so "future scaling can recommend substitutions" can't actually be exercised. Related decision (derive vs. hand-curate) raised on #1630.
- **Empty / collapsed function labels.** ~15 seeded movements get `function_roles: { primary: [] }` (Box Jump, Dumbbell Farmers Carry, Wall Walk, run/row/swim, Double-under…), making them invisible to `Movement.with_function`. And the 8-function vocabulary has no lunge/carry/jump/throw/pull-under, so lunges are folded into `squat`. Since #1630 consumes these labels directly, worth deciding intentionally which movements are legitimately function-less (pure mono) vs. just unmodeled.
- **Dropped `load-bearing metadata` work item.** #1629's Work list includes it; this PR deliberately omits it (documented rationale: infer from family + prescription, defer to #1684). Defensible and not required by the acceptance criteria — just flagging it's a conscious scope deviation for the epic owner to bless.

### Code-level (minor)

- `app/models/movement.rb` / `movement_function_assignable.rb`: `normalize_function_roles` is defined identically in both the concern's `class_methods` block and on `Movement`; since `include` runs first, Movement's copy shadows the concern's (dead code). The concern also hard-depends on `function_value`/`role_value`/`function_name`/`self::FUNCTIONS` that only exist on `Movement`, so despite its name it isn't a reusable mixin. Consolidate the helpers one place.
- `movement_function_assignable.rb` `after_save :sync_function_roles`: the guard `defined?(@function_role_assignments)` stays truthy for the record's lifetime, so any later `update` (e.g. renaming a movement) re-runs `delete_all` + bulk `create!` on the assignments even when roles didn't change. Idempotent but churns child rows; consider clearing the ivar after sync or diffing before rewriting.
- `Movement#family_movements` is unused (history goes through `for_movement_family`) and its comment refers to "history" though it returns movements. Either wire it in or drop it per AGENTS.md ("no speculative features").
- `MovementSubstitution` allows contradictory directed pairs (`A→B easier` and `B→A easier` can both exist). Not exercised yet; worth a validation or a documented invariant before the substitution engine consumes the graph.

### Sequencing

Epic order is #1628 → #1629 → #1630. Fine to land taxonomy first (independent), but #1628 (lazy `workout_instances`) needs to land before #1630 has anywhere to store generated recommendations.

Yanchek99:
Addressed the actionable code feedback in 6d6f1fe.

Changes made:
- Tightened movement fallback history when called with a Movement: it now filters by family + equipment + primary movement functions instead of returning the whole family. Passing a raw family still preserves the old family-only behavior.
- Consolidated the function-role helper methods into MovementFunctionAssignable and removed the duplicate Movement methods.
- Cleared the transient function-role assignment ivar after sync so later unrelated updates do not delete/recreate child assignment rows.
- Removed the unused Movement#family_movements helper and its speculative test.
- Added validation for contradictory inverse substitution directions.

I left the seed substitution graph, empty/unmodeled function labels, and load-bearing metadata as explicit design/scope follow-ups rather than guessing in this PR, since those points tie into #1630/#1684 and the earlier taxonomy decisions.

Verification:
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_log_test.rb test/models/movement_substitution_test.rb test/models/user_test.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel app/models/movement.rb app/models/movement_log.rb app/models/movement_substitution.rb app/models/concerns/movement_function_assignable.rb test/models/movement_test.rb test/models/movement_log_test.rb test/models/movement_substitution_test.rb test/models/user_test.rb
- git diff --check

Yanchek99:
Resolved the merge conflicts by merging origin/master into this branch in fb31e4e.

What changed:
- Kept the movement taxonomy migration ordered after the newer master migrations at 20260701120000.
- Preserved master’s ladder/team schema changes and Open/Hero seed additions.
- Resolved conflicts in cf/docs/decisions.md, db/schema.rb, and db/seeds.rb.
- Added the missing lateral_burpee_over_dumbbell seed local needed by the Open workout seeds.

Checks run:
- rvm 4.0.5@wod-tracker do ruby -c db/seeds.rb
- static seed movement local check across db/seeds/*.rb
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_log_test.rb test/models/movement_substitution_test.rb test/models/user_test.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel db/seeds.rb
- git diff --check
- git diff --cached --check

Yanchek99:
Addressed the naming feedback in a8b77b0.

Changes:
- Renamed MovementFunctionAssignment to MovementFunctionRole.
- Renamed MovementFunctionAssignable to MovementFunctionRoles.
- Renamed the table/association/fixture from movement_function_assignments to movement_function_roles.
- Updated the migration and schema so the PR creates the renamed table directly.

Checks run:
- rvm 4.0.5@wod-tracker do ruby -c app/models/concerns/movement_function_roles.rb
- rvm 4.0.5@wod-tracker do ruby -c app/models/movement_function_role.rb
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_log_test.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel app/models/movement.rb app/models/concerns/movement_function_roles.rb app/models/movement_function_role.rb app/models/movement_log.rb test/models/movement_test.rb
- git diff --check
- git diff --cached --check

Yanchek99:
Resolved the current merge conflicts in 8760008 by merging origin/master into the PR branch.

Details:
- The only content conflict was db/schema.rb.
- Master now has migrations at 20260701120000 and 20260701130000, so I bumped this PR’s movement taxonomy migration to 20260701140000.
- Updated db/schema.rb to version 2026_07_01_140000 while preserving master’s unit_system/load_unit schema changes and the PR’s movement_function_roles/movement_substitutions tables.

GitHub now reports the PR as mergeable; mergeStateStatus is still BLOCKED, which appears to be required checks/review rather than conflicts.

Checks run:
- rvm 4.0.5@wod-tracker do ruby -c db/migrate/20260701140000_add_movement_taxonomy_and_substitutions.rb
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_log_test.rb test/models/movement_substitution_test.rb test/models/user_test.rb
- git diff --check

Note: git diff --cached --check reported trailing whitespace in upstream CF WOD HTML fixture files added from master; I left those captured fixture files unchanged.

Yanchek99 (commented):
## Overview

This PR closes #1629 by adding a structured movement taxonomy (family, function role assignments, equipment, skill level) to `Movement`, directed movement substitutions, and history query helpers on `MovementLog`/`User`. It's a large but mostly mechanical change: two new models (`MovementFunctionRole`, `MovementSubstitution`), a shared `MovementFunctionRoles` concern, a migration/schema update, an extensive `db/seeds.rb` rewrite classifying all ~214 existing movements, and matching model/system tests. Docs (`cf/docs/decisions.md`, `cf/docs/movement-standards.md`) are updated to record the modeling rationale. Verified the seed rewrite preserves all 214 pre-existing movement names 1:1 with no drops or accidental duplicates.

## Correctness issue (see inline comment)

`Movement.supporting_implement_count` was not updated to match `supports_implement_count?`, which breaks the implement-count field toggle for equipment-classified movements whose name doesn't contain "dumbbell"/"kettlebell" (e.g. the seeded **Renegade Row**, `equipment: :dumbbell`). See the inline comment on `app/models/movement.rb` for details and a repro path through `implement_count_controller.js`.

## Other observations (minor, non-blocking)

See inline comments for specifics on:
- Seed re-run churn in `MovementFunctionRoles#sync_function_roles`.
- Whether blocking a same-direction reverse `lateral` substitution is intentional.
- Unrelated system-test changes bundled into this PR (`workout_builder_toolbar_test.rb`).

## Test coverage

Good coverage of the new models: uniqueness/self-reference/contradiction validations on `MovementSubstitution`, function-role querying and role-preservation-on-unrelated-update on `Movement`, and the three `MovementLog` history scopes. Gap: no test exercises `Movement.supporting_implement_count` / `implement_count_movement_ids` against an equipment-classified, non-name-matching movement — this is exactly what would have caught the bug flagged above.

## Style/conventions

Code follows existing Rails conventions well (enum usage, concern extraction, scope naming, fixture style). `db/migrate/20260701140000_add_movement_taxonomy_and_substitutions.rb` and `db/schema.rb` are in sync (version bumped correctly, indexes/foreign keys match).

## Security

No concerns — no new user input paths, no controller/params changes in this diff (taxonomy fields aren't yet exposed to end users, consistent with the PR's stated "future scaling work" framing).

app/models/movement.rb
```diff
@@ -1,17 +1,76 @@
...
+           inverse_of: :substitute_movement
+  has_many :substituted_for_movements, through: :inverse_movement_substitutions, source: :movement
+
+  enum :family, FAMILIES, prefix: true
+  enum :equipment, EQUIPMENT, prefix: true
+  enum :skill_level, SKILL_LEVELS, prefix: true
 
   validates :name, presence: true, uniqueness: { case_sensitive: false }
 
   scope :supporting_implement_count, -> { where('name ~* ?', IMPLEMENT_COUNT_NAME_PATTERN.source) }
```
Yanchek99: **Bug:** this scope wasn't updated to match `supports_implement_count?` below, which now also checks `equipment_dumbbell?`/`equipment_kettlebell?`. The two are now inconsistent.

This scope feeds `Movement.supporting_implement_count.ids` in `app/helpers/application_helper.rb:14` (`implement_count_movement_ids`), which is serialized to the DOM and read by `app/javascript/controllers/implement_count_controller.js` to toggle the implement-count field client-side when the movement dropdown changes. That JS file's own comment says: "swap that source for the equipment taxonomy once it lands" — it wasn't swapped.

Concretely, this PR's own seed data reproduces the bug: **Renegade Row** is seeded with `equipment: :dumbbell` but its name doesn't match `/dumbbell|kettlebell/i`. Initial server render shows the field correctly (via `supports_implement_count?`), but selecting Renegade Row from the movement dropdown client-side won't find its id in `movement_ids`, so the field will incorrectly stay hidden — and `toggle(event)` will call `clearInput()`, wiping any value the user already entered.

Suggest:
```ruby
scope :supporting_implement_count, -> {
  equipment_dumbbell.or(equipment_kettlebell).or(where('name ~* ?', IMPLEMENT_COUNT_NAME_PATTERN.source))
}
```
and a test asserting the scope includes a non-name-matching, equipment-classified movement (e.g. Renegade Row) to lock in parity.

app/models/movement.rb
```diff
@@ -1,17 +1,76 @@
...
 
   def supports_implement_count?
-    name.to_s.match?(IMPLEMENT_COUNT_NAME_PATTERN)
+    equipment_dumbbell? || equipment_kettlebell? || name.to_s.match?(IMPLEMENT_COUNT_NAME_PATTERN)
```
Yanchek99, line 78: This predicate now diverges from the `supporting_implement_count` scope at line 70 — see the comment there for the concrete failure case (Renegade Row) and suggested fix.

app/models/concerns/movement_function_roles.rb
```diff
@@ -0,0 +1,82 @@
...
+                 role: role_value(role)
+               })
+        .distinct
+    }
+
+    self::FUNCTIONS.each_key do |function_name|
+      define_method(:"function_#{function_name}?") { function?(function_name) }
+    end
+
+    after_save :sync_function_roles, if: :function_roles_assigned?
```
Yanchek99, line 25: Minor efficiency note: `function_roles_assigned?` is gated on whether `function_roles=` was ever invoked, not on whether the assignments actually changed. `db/seeds.rb` calls `assign_attributes(function_roles: {...})` unconditionally for every movement on every run, so re-running seeds against an already-seeded DB will `delete_all` + recreate every `MovementFunctionRole` row for every movement each time, churning surrogate IDs even when nothing changed. Not incorrect, but worth avoiding if seeds ever run repeatedly against a long-lived environment.

app/models/movement_substitution.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  private
+
+  def substitute_is_different_movement
+    return if movement_id.blank? || substitute_movement_id.blank?
+
+    errors.add(:substitute_movement, 'must be different from movement') if movement_id == substitute_movement_id
+  end
+
+  def inverse_direction_is_consistent
```
Yanchek99: Worth double-checking the semantics here for `lateral`: this blocks a reverse row with the *same* direction as the forward row. That's clearly correct for `easier`/`harder` (mutually exclusive claims — A can't be both easier and harder than B). But for `lateral`, a same-direction reverse entry isn't logically contradictory, just redundant (since `substitutes`/`substituted_for_movements` already expose both directions from a single row). If blocking the redundant insert is intentional, might be worth a comment; if not, this could reject a legitimate explicit `lateral` entry in the reverse direction.

test/system/workout_builder_toolbar_test.rb
```diff
@@ -14,7 +14,7 @@ class WorkoutBuilderToolbarTest < ApplicationSystemTestCase
   test 'keeps workout builder toolbar fixed to the bottom while scrolling' do
     visit new_workout_url
 
-    8.times { click_on 'Add Exercise', match: :first }
+    8.times { click_on 'Add Segment' }
```
Yanchek99: This change (`'Add Exercise'` → `'Add Segment'`) looks unrelated to the movement taxonomy work in this PR. Per the repo's scoping conventions (avoid unrelated changes bundled into a feature PR), consider splitting this into its own PR or explaining the connection in the PR description.

Yanchek99:
Brought this branch up to date with current master in 1e84f16.

Resolved conflicts in:
- cf/docs/decisions.md
- db/schema.rb
- test/fixtures/movements.yml
- test/system/exercise_card_summary_formatting_test.rb

Also bumped the movement taxonomy migration to 20260721120000 so it sorts after master’s latest migrations. While verifying the conflicted system test, I fixed the merged local-summary behavior so a 0-calorie prescription still renders as max calories Row after a local card save.

Local checks run:
- rvm 4.0.5@wod-tracker do ruby -c db/migrate/20260721120000_add_movement_taxonomy_and_substitutions.rb
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_log_test.rb test/models/movement_substitution_test.rb test/services/cf_wod/movement_lookup_test.rb
- yarn build
- rvm 4.0.5@wod-tracker do bundle exec rails test test/system/exercise_card_summary_formatting_test.rb
- git diff --check
- git diff --cached --check

GitHub currently reports the PR as mergeable; required checks are still running.

Yanchek99:
Updated the seed taxonomy so AbMat Sit-up is classified as primary trunk flexion instead of hinge. Verified with `ruby -c db/seeds.rb`, `bundle exec rubocop db/seeds.rb`, and `git diff --check`.

Yanchek99:
I agree that `squat` + `vertical_push` was the wrong mapping for Burpee Box Jump. `vertical_push` implies an overhead press, and the box-jump phase does not map cleanly to the sourced function vocabulary, so I changed Burpee Box Jump and the related burpee-over variants to use `horizontal_push` for the burpee floor-press phase. Verified with `ruby -c db/seeds.rb`, `bundle exec rubocop db/seeds.rb`, and `git diff --check`.

Yanchek99:
Took another pass over the seeded function assignments and pushed `f764111`.

Updated clear mismatches:
- Body Blaster: `horizontal_push` + `vertical_pull` instead of squat/vertical push/pull
- Dumbbell Turkish Get-up: blank function roles because the current vocabulary does not cleanly describe the get-up sequence
- Shoot-through: blank function roles instead of squat/vertical push
- Slam Ball: hinge only, removed vertical push
- Sled Pull: horizontal pull instead of vertical pull
- Sumo Deadlift High Pull: hinge primary, vertical pull secondary

Also checked migration ordering against current `origin/master`; master still ends at `20260716180000`, and this PR migration is `20260721120000`, so it is ordered after master.

Verified with:
- `ruby -c db/seeds.rb`
- `ruby -c db/migrate/20260721120000_add_movement_taxonomy_and_substitutions.rb`
- `bundle exec rubocop db/seeds.rb`
- `bundle exec rubocop --parallel`
- `bundle exec rails test test/models/movement_test.rb test/models/movement_log_test.rb test/models/movement_substitution_test.rb`
- `bundle exec rails test test/services/cf_wod/movement_lookup_test.rb`
- `git diff --check`

Yanchek99:
Updated in `ac4c5df` to broaden Olympic lift function roles.

Applied the squat-clean model across the taxonomized clean/snatch variants: power, hang-power, split, and hang snatch/clean rows now keep `hinge` + `squat` as primary where there is a receiving/catch phase, with `vertical_pull` as secondary. Clean-and-jerk / ground-to-overhead variants now keep `hinge` + `squat` + `vertical_push` primary and `vertical_pull` secondary, so the jerk/overhead portion is still represented without dropping the squat/catch function.

I left later Hero seed rows that still use plain `find_or_create_by` without taxonomy untouched rather than inventing new metadata in this pass.

Verified with:
- `ruby -c db/seeds.rb`
- `bundle exec rubocop db/seeds.rb`
- `bundle exec rails test test/models/movement_test.rb test/models/movement_log_test.rb test/models/movement_substitution_test.rb test/services/cf_wod/movement_lookup_test.rb`
- `bundle exec rubocop --parallel`
- `git diff --check`

Yanchek99:
Fixed in `fa4de58`.

- Renamed the seeded canonical movement from `Clean Squat` / `clean_squat` to `Squat Clean` / `squat_clean` and updated hero workout seed references.
- Updated `Squat Clean` function roles to primary `hinge` + `squat`, secondary `vertical_pull`.
- Added taxonomy for the related squat-clean rows: `Dumbbell Hang Squat Clean`, `Dumbbell Squat Clean`, `Hang Squat Clean`, `Squat Clean Thruster`, and `Dumbbell Squat Clean Thruster`.
- Added `Squat Snatch` taxonomy with primary `hinge` + `squat`, secondary `vertical_pull` + `vertical_push`.

Verification run after the seed changes:
- `ruby -c db/seeds.rb`
- `ruby -c db/seeds/hero_workouts.rb`
- `bundle exec rubocop db/seeds.rb db/seeds/hero_workouts.rb`
- `bundle exec rails test test/models/movement_test.rb test/models/movement_log_test.rb test/models/movement_substitution_test.rb test/services/cf_wod/movement_lookup_test.rb`
- `bundle exec rubocop --parallel`
- `git diff --check`

Yanchek99:
Addressed the seed taxonomy coverage feedback in ea78fcd.

What changed:
- Converted the remaining Hero and partner/team Hero movement seeds from bare Movement.find_or_create_by calls to taxonomy-aware Movement.find_or_initialize_by(...).tap blocks.
- Added family/equipment/skill taxonomy for the omitted seeded movements, including Bike, Bear Crawl, Broad Jump, carries, burpee + pull variants, sled work, shuttle run, triple-under, waiters walk, etc.
- Kept function_roles: { primary: [] } for movements where the current sourced function vocabulary does not cleanly apply, such as Bike/Bear Crawl/carries/holds/static inversion-style movements, instead of inventing unsupported functions.
- The earlier Squat Clean rename and Squat Snatch seed remain in this PR.

Verification run:
- rvm 4.0.5@wod-tracker do ruby -c db/seeds.rb: Syntax OK
- rvm 4.0.5@wod-tracker do ruby -c db/seeds/hero_workouts.rb: Syntax OK
- rvm 4.0.5@wod-tracker do bundle exec rubocop db/seeds.rb db/seeds/hero_workouts.rb: 2 files inspected, no offenses
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_log_test.rb test/models/movement_substitution_test.rb test/services/cf_wod/movement_lookup_test.rb: 24 runs, 80 assertions, 0 failures
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel: 168 files inspected, no offenses
- git diff --check: no whitespace errors

Yanchek99:
Brought the PR back up to date with current master in 541489c.

What changed:
- Merged origin/master into codex/1629-movement-taxonomy.
- No textual conflicts during the merge.
- Confirmed the taxonomy migration still sorts after the latest master migrations: master is still before 20260721120000_add_movement_taxonomy_and_substitutions.rb.
- No unresolved inline review threads are currently open; GitHub was reporting the PR as BEHIND before this merge.

Local verification run after the merge:
- rvm 4.0.5@wod-tracker do ruby -c db/migrate/20260721120000_add_movement_taxonomy_and_substitutions.rb: Syntax OK
- rvm 4.0.5@wod-tracker do ruby -c db/seeds.rb: Syntax OK
- rvm 4.0.5@wod-tracker do ruby -c db/seeds/open_workouts.rb: Syntax OK
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_log_test.rb test/models/movement_substitution_test.rb test/services/cf_wod/movement_lookup_test.rb test/models/log_test.rb test/controllers/logs_controller_test.rb: 56 runs, 207 assertions, 0 failures
- yarn build: completed
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel: 169 files inspected, no offenses
- git diff --check: no whitespace errors

GitHub checks are running again on the updated branch.

Yanchek99:
## Code review — movement taxonomy and substitutions

Read-only pass over `8efdee2..fa4de58` (merge-base with `master` → head). CI is green (Build, rubocop, CodeQL), so this focuses on design, data safety, and domain correctness rather than reproducing green checks. Seeded taxonomy was spot-checked against `cf/docs/movement-standards.md` and `cf/docs/OVERVIEW.md`.

**Verdict: ready to merge with fixes.** No Critical issues. One Important item has production-data impact and should be resolved before merge.

### Strengths

- **Schema design is solid.** Scalar taxonomy (`family`, `equipment`, `skill_level`) as indexed integer enums on `movements`, with functions/roles correctly extracted to a `movement_function_roles` join table (many functions per movement, each with a role). FKs, the unique composite index `idx_movement_function_roles_unique` on `[movement_id, movement_function]`, and `idx_movement_substitutions_unique_pair` all exist at the DB level *and* are mirrored by model validations. The migration is `change`-reversible and `db/schema.rb` matches exactly (version `2026_07_21_120000`).
- **Domain-faithful taxonomy.** Spot-checked ~15 seeded movements: Deadlift → hinge; Snatch/Clean → hinge + squat primary with vertical_pull secondary; Thruster → squat primary + vertical_push secondary; Overhead Squat → advanced; GHD Sit-up → trunk_flexion; Double-under → monostructural with no function. The primary/secondary/tertiary role weighting for compound lifts is a genuinely good modeling call, and the rationale (including why no load boolean is stored) is documented in `cf/docs/decisions.md`.
- **Taxonomy seeding is idempotent.** The `find_or_initialize_by(...).tap { assign_attributes; save! }` pattern plus the concern's `function_role_assignments_unchanged?` guard means re-running seeds does not churn function-role rows — covered by the "does not rewrite function roles" tests in `test/models/movement_test.rb`.
- **Empty-taxonomy edge case handled.** `function_roles: { primary: [] }` (Box Jump, sleds) normalizes to no rows without error.
- Substitution self-reference and duplicate-pair validations are present and tested.

### Issues

#### Important

**1. The `Clean Squat` → `Squat Clean` rename orphans existing production data.**

`db/seeds.rb:124` now seeds `name: 'Squat Clean'` (previously `'Clean Squat'`), and `db/seeds/hero_workouts.rb` was updated to reference the new variable in ~15 hero workouts. A name diff confirms this is the only rename in the file.

Because those hero `Workout` records already exist in production, `Workout.find_or_create_by(name: ...)` will find and skip them — so their exercises keep pointing at the old, now-untaxonomied `Clean Squat` movement, while `find_or_initialize_by(name: 'Squat Clean')` creates a new taxonomied-but-unreferenced duplicate.

Net production result: two clean-squat movements, and every existing Squat Clean hero WOD loses the taxonomy benefit this PR exists to deliver.

*Fix:* add a data migration that renames the existing `Clean Squat` record to `Squat Clean` (or otherwise reconciles the two), rather than relying on a seed re-run.

**2. `MovementLog.for_movement_family` has two materially different behaviors behind one name.**

In `app/models/movement_log.rb`, passing a bare family (`:gymnastics`) filters on family only; passing a `Movement` filters on family **and** equipment **and** matching primary functions.

The method name — and the movement-family fallback acceptance criterion — imply family-level matching, but the `Movement` path is much narrower. A barbell Thruster will not match a dumbbell Thruster (different equipment), and any movement with a differing primary function set is excluded. That is a reasonable "similar movement" query, but it is mislabeled as "family," and the divergence between the two call signatures is surprising for callers.

*Fix:* either rename (e.g. `similar_to`) or document the two contracts explicitly — and reconsider whether equipment should gate a *family* fallback at all.

**3. The "single directed row per pair" invariant is documented but not enforced.**

In `app/models/movement_substitution.rb`, `inverse_direction_is_consistent` only rejects a reverse row with the *same* direction. `(A→B easier)` plus `(B→A harder)` are both allowed, since the unique index is on the ordered `[movement_id, substitute_movement_id]` — yet the inline comment claims "a single directed row owns each movement pair."

So redundant reversed rows can coexist, and there is no helper that unions `substitutes` + `substituted_for_movements` while inverting easier/harder. A future "recommend substitutions" consumer would have to traverse both associations and flip direction by hand.

*Fix:* either enforce the unordered-pair invariant (normalize the pair on write, or forbid any reverse row) or drop the "single row owns the pair" claim and provide a bidirectional lookup helper. Separately, flagging a reverse *lateral* row as "contradicts inverse substitution" is a misleading message — a reverse lateral is redundant, not contradictory.

#### Minor

**4. No substitutions are seeded.** 159 movements get taxonomy, but `db/seeds.rb` creates zero `MovementSubstitution` records (only two fixtures exist). Reasonable for a foundation PR, but the substitution-recommendation framing is currently unexercised by real data — worth a follow-up issue.

**5. Debatable family assignments for load inference.** `Wall-ball Shot` and `Sled Drag`/`Sled Pull` are seeded `family: :weightlifting`. Per the decision doc this is deliberate (load is inferred from the weightlifting family), but none of these is a weightlifting-modality movement in the strict monostructural/gymnastics/weightlifting sense. Flagging to confirm intent.

**6. Minor query inefficiency.** In `for_movement_family`, the `Movement` path chains `joins(:movement)` and then `joins(movement: :movement_function_roles)`, relying on `.distinct` to dedupe; a single join expression would be cleaner. `movement.movement_function_roles.role_primary.pluck(...)` also fires an extra query per call — negligible, and there is no N+1 across log rows.

**7. `function_roles=` writes are validated only at `after_save` `create!` time**, not during movement validation, so invalid input raises mid-save instead of surfacing as a validation error. No UI form consumes this yet, so impact is low today — it will matter once one does.

### Recommendations

- Add the rename data migration (issue 1) before merge; it is the only item with production data impact.
- Clarify the `for_movement_family` contract (issue 2), since it directly backs an acceptance criterion.
- Add a substitution model test for the reversed-but-consistent case (`A→B easier` + `B→A harder`) to pin down the intended pair invariant (issue 3).

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

Yanchek99:
Addressed the important review items in 759d0f0.

Changes made:
- Added migration-level reconciliation for Clean Squat -> Squat Clean. If only Clean Squat exists, the migration renames it. If both names exist, it reassigns exercises and movement logs to Squat Clean before removing the duplicate Clean Squat row. This avoids production hero workouts staying attached to an untaxonomied legacy movement.
- Split the history query contracts. MovementLog.for_movement_family now remains family-only even when passed a Movement; the narrower family + equipment + primary-function fallback is now MovementLog.for_similar_movement, exposed on User as similar_movement_history_for.
- Enforced the substitution pair invariant both in the model and DB. Reverse rows are rejected regardless of direction, and the migration/schema now use an unordered expression unique index over LEAST/GREATEST movement ids. Added tests for same-direction, complementary-direction, lateral reverse rows, and DB-level validation bypass.

Verification run:
- rvm 4.0.5@wod-tracker do ruby -c db/migrate/20260721120000_add_movement_taxonomy_and_substitutions.rb: Syntax OK
- rvm 4.0.5@wod-tracker do bundle exec rails db:test:prepare: completed
- rvm 4.0.5@wod-tracker do bundle exec rails test test/models/movement_test.rb test/models/movement_log_test.rb test/models/movement_substitution_test.rb test/models/user_test.rb test/services/cf_wod/movement_lookup_test.rb: 33 runs, 121 assertions, 0 failures
- rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel: 169 files inspected, no offenses
- rvm 4.0.5@wod-tracker do bundle exec rails runner migration SQL helper smoke check: completed
- git diff --check: no whitespace errors

db/seeds.rb
```diff
@@ -15,240 +15,1241 @@
...
+jumping_lunge = Movement.find_or_initialize_by(name: 'Jumping Lunge').tap do |movement|
+  movement.assign_attributes(family: :gymnastics, function_roles: { primary: [:squat] }, skill_level: :basic)
+  movement.save!
+end
+kettlebell_snatch = Movement.find_or_initialize_by(name: 'Kettlebell Snatch').tap do |movement|
+  movement.assign_attributes(family: :weightlifting, function_roles: { primary: [:hinge] },
+                             equipment: :kettlebell, skill_level: :advanced)
+  movement.save!
+end
+kettlebell_sumo_hi_pull = Movement.find_or_initialize_by(name: 'Kettlebell Sumo Hi Pull').tap do |movement|
```
Yanchek99, line 389: **Kettlebell Sumo Hi Pull** is missing the `:hinge` primary function role. Its barbell counterpart, `Sumo Deadlift High Pull` (line 728), correctly tags `primary: [:hinge], secondary: [:vertical_pull]` — same hip-hinge-into-upright-pull pattern, just a different implement. This entry only has `primary: [:vertical_pull]`, dropping the hinge phase entirely.

Suggest aligning it with the barbell version: `primary: [:hinge], secondary: [:vertical_pull]`.

db/seeds.rb
```diff
@@ -15,240 +15,1241 @@
...
+  movement.assign_attributes(family: :weightlifting, function_roles: { primary: [] }, equipment: :sled,
+                             skill_level: :basic)
+  movement.save!
+end
+sled_pull = Movement.find_or_initialize_by(name: 'Sled Pull').tap do |movement|
+  movement.assign_attributes(family: :weightlifting, function_roles: { primary: [:horizontal_pull] }, equipment: :sled,
+                             skill_level: :basic)
+  movement.save!
+end
+snatch = Movement.find_or_initialize_by(name: 'Snatch').tap do |movement|
```
Yanchek99, line 637: Question on the Snatch family's function-role tagging: `Snatch` here (and `Power Snatch` L511, `Squat Snatch` L1045, `Hang Snatch` L354, `Split Snatch` L666, `Hang Squat Snatch` L1202, `3-Position Power Snatch` L1128, and the Dumbbell equivalents at L220, L236, L903, L1176) all tag `secondary: [:vertical_push]` for the overhead catch.

`cf/docs/programming.md` (pre-existing, not touched by this PR) describes the analogous phase of the Clean as "hinge/pull with a **receiving position**" — explicitly distinguishing a catch from an active press. None of the plain Clean entries in this file tag `vertical_push` (correctly — the push only appears once a Jerk is added, e.g. `Clean and Jerk` L109). By that same logic, shouldn't the Snatch's overhead catch also be a receive rather than a push, i.e. drop `vertical_push` here too?

This is also inconsistent with `Kettlebell Snatch` (L384) added in this same PR, which is tagged `primary: [:hinge]` only — no pull, no push, no squat — treating the KB snatch's catch very differently from the barbell version. Worth reconciling toward one consistent model, or confirming the difference is intentional.

db/seeds.rb
```diff
@@ -15,240 +15,1241 @@
...
+  movement.save!
+end
+muscle_snatch = Movement.find_or_initialize_by(name: 'Muscle Snatch').tap do |movement|
+  movement.assign_attributes(family: :weightlifting,
+                             function_roles: { primary: [:hinge], secondary: [:vertical_pull, :vertical_push] },
+                             equipment: :barbell,
+                             skill_level: :advanced)
+  movement.save!
+end
+muscle_up = Movement.find_or_initialize_by(name: 'Muscle-up').tap do |movement|
```
Yanchek99, line 471: Minor question: `Muscle-up` here (and `Bar Muscle-up` L39, `Ring Muscle-up` L561, `Strict Bar Muscle-up` L679, `Strict Muscle-up` L703) are tagged `primary: [:vertical_pull]` only. A muscle-up is generally coached as a pull (like a pull-up) into a dip/transition (a push) — structurally similar to how `Clean and Jerk` gets both `vertical_pull` and `vertical_push` tagged. Should these carry `secondary: [:vertical_push]` for the transition, or is the pull considered dominant enough that the push isn't worth tagging? Not blocking, just flagging for consistency with the rest of the taxonomy.

Yanchek99 (commented):
Reviewed with special attention to `db/seeds.rb` movement categorization, per request. The taxonomy/substitution modeling itself (migration, sibling-aware uniqueness constraints, the `MovementFunctionRoles` concern's diff-and-skip sync, the history query scopes) is solid and well-tested — I traced through the sync/uniqueness logic and the tests exercise the right edge cases.

Found a handful of movement-categorization inconsistencies worth confirming, plus two smaller consistency/scope notes — left inline below. None of these look blocking; mostly flagging tagging choices that seem internally inconsistent with either another entry in this same file or with `cf/docs/programming.md` (pre-existing, not touched by this PR), so you can confirm intent.

Note: I wasn't able to run the test suite myself in this environment (RVM/gemset didn't initialize in the sandbox), so I'm relying on the PR description's stated validation (`test/models`, `rubocop --parallel`) rather than re-verifying green CI myself.

app/models/movement.rb
```diff
@@ -1,17 +1,78 @@
...
+    hinge: 1,
+    vertical_push: 2,
+    vertical_pull: 3,
+    horizontal_push: 4,
+    horizontal_pull: 5,
+    trunk_flexion: 6,
+    trunk_extension: 7
+  }.freeze
+
+  EQUIPMENT = {
```
Yanchek99, line 24: `cf/docs/movement-standards.md` (added by this PR) explicitly says loaded movements "can prescribe barbell, dumbbell, kettlebell, medicine ball, vest, sled, sandbag, or object load" — but this `EQUIPMENT` enum has no `sandbag` or `vest` value. That leaves `Sandbag Carry` (seeds.rb:1019), `Sandbag Clean` (seeds.rb:1023), and `Plate Carry` (seeds.rb:1010) with no equipment classification even though their primary implement is unambiguous.

Was `sandbag`/`vest` intentionally left out of the taxonomy (e.g. too rare to justify a dedicated bucket vs. `mixed`), or is this a gap for a follow-up?

test/fixtures/movements.yml
```diff
@@ -1,30 +1,58 @@
 # Read about fixtures at http://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
 hspu:
   name: Handstand Push Up
+  family: 1
```
Yanchek99: These new taxonomy fixtures (here, and in `movement_function_roles.yml` / `movement_substitutions.yml`) use raw enum integers (`family: 1`, `equipment: 5`, `movement_function: 2`, `role: 0`, `direction: 0`, etc.) instead of the symbolic labels the rest of the fixture suite uses for enum columns — e.g. `test/fixtures/workouts.yml` uses `score_type: time` and `test/fixtures/users.yml` uses `sex: female`. Rails fixtures accept enum string labels directly, so this should work the same way here: `family: gymnastics`, `equipment: pull_up_bar`, `movement_function: vertical_push`, `role: primary`, `direction: easier`, etc.

I checked the numeric values against `Movement::FAMILIES`/`FUNCTIONS`/`EQUIPMENT`/`FUNCTION_ROLES` and `MovementSubstitution::DIRECTIONS` and they're all correct — this is a readability/consistency ask, not a data bug. Right now every value here requires cross-referencing those hashes to know what it means.

app/javascript/controllers/exercise_card_controller.js
```diff
@@ -428,6 +428,7 @@ class LeadingPrescription {
     if (!this.structuralSingleRepMetric(candidate)) return true
 
     return !this.metrics.some((metric) => (
+      this.maxCalorie(metric) ||
```
Yanchek99: This adds `this.maxCalorie(metric) ||` to `repCanLead`'s exclusion check — a behavioral change to leading-prescription selection that doesn't look related to movement taxonomy/substitutions, and isn't mentioned in the PR summary or validation notes. Intentional fix bundled in here, or a rebase/merge leftover? If intentional, might be worth its own PR with a test covering the max-calorie case it addresses.

db/seeds.rb
Yanchek99: Strengthening this with a concrete rule rather than just doc interpretation: in competitive Olympic weightlifting, an athlete who visibly presses out the bar during the overhead lockout after the catch receives a **no-lift** — a "press-out" is an explicit foul under IWF/USAW rules. That's about as authoritative as it gets that the catch is a *receive*, not a press: there is no legal pressing action in a snatch's overhead phase at all.\n\nThat also applies to `Devil Press` (seeds.rb:1155, tagged `secondary: [:vertical_push]`) — CrossFit's own programming notes describe it as a burpee into a two-dumbbell hang-power-snatch-style pull to overhead, i.e. the same hip-driven turnover as a barbell snatch, not a press. Given that, I'd upgrade this from a question to a fairly confident ask: drop `vertical_push` from the Snatch family (and Devil Press), unless there's a reason specific to this app's model that I'm missing.

db/seeds.rb
Yanchek99: Applying more gymnastics-specific coaching knowledge: this is a well-established two-phase breakdown, not really in dispute. A muscle-up is universally coached as a pull (getting the hips/chest above the rings or bar) followed by a **dip** to full lockout in support — the dip is the same pressing pattern as a ring dip or bar dip, just under momentum from the pull. It's the same reason a false-grip pull-up alone doesn't equal a muscle-up: without the dip/transition, the athlete never reaches support.\n\nGiven that, and given this file already tags both phases for Clean-and-Jerk-style movements, I'd lean towards this being a real gap rather than a judgment call: `secondary: [:vertical_push]` seems warranted on `Muscle-up`, `Bar Muscle-up`, `Ring Muscle-up`, `Strict Bar Muscle-up`, and `Strict Muscle-up`.

db/seeds.rb
```diff
@@ -15,240 +15,1241 @@
...
+                             skill_level: :intermediate)
+  movement.save!
+end
+overhead_dumbbell_lunge = Movement.find_or_initialize_by(name: 'Overhead Dumbbell Lunge').tap do |movement|
+  movement.assign_attributes(family: :weightlifting, function_roles: { primary: [:squat] },
+                             equipment: :dumbbell,
+                             skill_level: :intermediate)
+  movement.save!
+end
+overhead_squat = Movement.find_or_initialize_by(name: 'Overhead Squat').tap do |movement|
```
Yanchek99, line 492: New finding, using the project's own cited domain source: `cf/docs/terminology.md` (pre-existing, not touched by this PR) states "The L1 Course teaches nine foundational movements — the air squat, front squat, **overhead squat**, shoulder press, push press, push jerk, deadlift, sumo deadlift high pull, and medicine-ball clean." Every other movement in that list of nine is tagged `basic` (Air Squat, Front Squat, Shoulder Press, Push Press, Deadlift, Medicine-ball Clean) or `intermediate` (Push Jerk L528, Sumo Deadlift High Pull L728 — the two most technically demanding of the nine). Overhead Squat is the only one of the nine tagged `advanced`, a full tier above even Push Jerk/SDHP.\n\nGiven CrossFit's own L1 curriculum teaches it as a foundational movement in the first weeks of training (same tier as air squat and deadlift), this looks miscategorized — `intermediate` would match the pattern set by the other technically-demanding foundational lifts, or `basic` would match the rest of the nine. `advanced` seems to overstate it relative to the app's own precedent.

Yanchek99: Retracting this one. I conflated "taught as one of the L1 foundational movements" with "should be a low/basic difficulty tier" — those are different axes. Foundational just means it's part of the core nine every athlete learns early; it says nothing about how demanding the movement is to execute well.

In practice the overhead squat is generally regarded as the *most* technically demanding of the nine — it requires shoulder flexion/external rotation, thoracic extension, ankle dorsiflexion, and hip mobility simultaneously while balancing a loaded bar overhead through a full squat, and it's commonly used as the single best test of an athlete's overall mobility. `advanced` is a reasonable, likely correct, categorization here. Sorry for the noise.

PR #1692, opened 2026-06-22
[codex] Seed CrossFit 260622

Yanchek99:
## Summary

- seed CrossFit daily workout 260622 from https://www.crossfit.com/260622
- add missing source movement names for freestanding shoulder taps and deficit push-ups
- render segmented total-rep clocks as `On a 20-minute clock for total reps`
- render named max-rep timed segments with their time windows

## Modeling

This workout is four 5-minute windows scored by total reps. Each window starts with a 200-meter run and then max reps of one gymnastics movement. The existing segment model can represent the structure, but the old objective text rendered these workouts like AMRAP rounds. This PR adds a narrow rendering path for segmented total-rep clocks so the seeded workout does not display misleadingly.

## Validation

- `rvm 4.0.5@wod-tracker do bundle exec rails test test/helpers/workouts_helper_test.rb test/helpers/segments_helper_test.rb`
- `rvm 4.0.5@wod-tracker do bundle exec ruby -c db/seeds.rb`
- `rvm 4.0.5@wod-tracker do bundle exec rails runner "load Rails.root.join('db/seeds.rb'); ..."`
- `git diff --check`

The Rails runner confirmed: `CrossFit 260622 | rep | 20 | 4 | 8 | 2026-06-22`.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -165,9 +165,6 @@ Movement.find_or_create_by(name: 'Zercher Squat')
 # Programs
 cfj = Program.find_or_create_by(name: 'Crossfit Journal')
 
-# ==============================================================================
-# CrossFit Daily Workouts
-# ==============================================================================
 # 260622
 # On a 20-minute clock for total reps:
 # 0:00-5:00:
@@ -188,7 +185,7 @@ cfj = Program.find_or_create_by(name: 'Crossfit Journal')
 #
 # Female 2-inch deficit
 # Male 4-inch deficit
-crossfit_260622 = Workout.find_or_create_by(name: 'CrossFit 260622') do |workout|
+crossfit_260622 = Workout.find_or_create_by(name: 'CF-260622') do |workout|
   workout.time = 20
   workout.score_type = :rep
   workout.notes = 'Each 5-minute window begins with a 200-meter run, followed by max reps of the gymnastics movement. '\
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -60,6 +60,7 @@ Movement.find_or_create_by(name: 'Dumbbell Thruster')
 Movement.find_or_create_by(name: 'Dumbbell Turkish Get-up')
 Movement.find_or_create_by(name: 'Forward Roll From Support')
 Movement.find_or_create_by(name: 'Freestanding Handstand Push-up')
+freestanding_shoulder_tap = Movement.find_or_create_by(name: 'Freestanding Shoulder Tap')
 Movement.find_or_create_by(name: 'Front Scale')
 Movement.find_or_create_by(name: 'Front Squat')
 Movement.find_or_create_by(name: 'GHD Back Extension')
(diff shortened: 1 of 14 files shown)


db/seeds.rb
```diff
@@ -163,6 +165,63 @@
...
+# 200-meter run
+# Max L pull-ups
+#
+# 15:00-20:00:
+# 200-meter run
+# Max deficit push-ups
+#
+# Female 2-inch deficit
+# Male 4-inch deficit
+crossfit_260622 = Workout.find_or_create_by(name: 'CrossFit 260622') do |workout|
```
Yanchek99:
```suggestion
crossfit_260622 = Workout.find_or_create_by(name: 'CF-260622') do |workout|
```

db/seeds.rb
```diff
@@ -163,6 +165,63 @@
 # Programs
 cfj = Program.find_or_create_by(name: 'Crossfit Journal')
 
+# ==============================================================================
+# CrossFit Daily Workouts
+# ==============================================================================
```
Yanchek99: Delete
```suggestion
```

Yanchek99 (commented):
## Code review

Well-scoped, well-tested PR. Seeds CrossFit 260622 (four 5-minute total-rep windows) and adds a rendering path so segmented total-rep clocks no longer display like AMRAP rounds. Notes below; only #1 is worth genuinely considering before merge.

### Correctness ✅
- **Predicate ordering is sound.** `max_reps?` is a strict subset of `amrap?` (it calls `amrap?` internally), and `timed_segment_prescription` checks `max_reps?` before `amrap?`, so the new branch wins. Same for `segmented_total_reps?` placed before `amrap?` in `workout_objective`.
- **Server/client parity is good.** Both sides gate on `time present + no rounds + no interval` (the JS reaches `hasMaxRepExercise` only after early-returning on interval/rounds), both use reps `== 0` / `"0"` as the max-rep sentinel, both trim before comparing, and the summary strings match exactly (`"<name>: max reps in 5 minutes"`).
- **`then_prefix` suppression reads well.** Suppressing `"Then, "` for named max-rep windows avoids `"Then, 0:00-5:00: ..."`, and the new `prefixes subsequent named segments with then` test guards the non-max-rep branch.

### Worth a look
1. **`segmented_total_reps?` is vacuously true with zero exercises.** `exercises.none? { |e| e.segment.blank? }` returns `true` when there are no exercises at all, so a rep-scored, timed workout with segments but no exercises would render "On a N-minute clock for total reps." Unusual/invalid state, but adding `exercises.any?` alongside `segments.any?` would make it strict. (`app/models/workout.rb`)
2. **`reps == 0` as the "max reps" sentinel is implicit.** Consistent across both code paths and matches the existing encoding, so no change needed — but a one-line comment on `Segment#max_reps?` noting that `reps: 0` means "max" would help the next reader. (`app/models/segment.rb`)

### Conventions ✅
Helper extraction follows the existing guard-clause style and stays `private`; the seed follows the established `find_or_create_by` + named-movement-variable pattern and the schedule line matches existing usage. New movements and the promotion of `Skin the Cat`/`L Pull-up` to locals are minimal and reasonable.

### Tests
Good coverage: helper unit tests for the workout objective and named max-rep segment, the negative `then`-prefix case, and a system test confirming the summary survives an expand/collapse round-trip (the actual bug class). Fixtures model the new state cleanly.

One gap: no **direct** model test for `Workout#segmented_total_reps?` / `Segment#max_reps?` — exercised only indirectly via the helpers. A direct unit test would pin the semantics, especially the zero-exercise edge in #1. Optional.

### Security / performance
No security surface (seed data + display helpers; notes go through `has_rich_text`). No perf concern — `Segment.default_scope` already `includes(:exercises)`, so `max_reps?`'s `exercises.any?` adds no queries in list rendering.

**Verdict:** Solid, low-risk, ready to merge. Only #1 is worth considering before merge; #2 and the model-test gap are nice-to-haves.

PR #1694, opened 2026-06-23
[codex] Split workout seeds

Yanchek99:
## Summary
- load workout seeds from dedicated benchmark, hero, and CF seed files
- keep users, movements, and programs in the top-level seed file
- remove obsolete movement local assignments from db/seeds.rb after the split

## Validation
- ruby -c db/seeds.rb db/seeds/benchmark_workouts.rb db/seeds/hero_workouts.rb db/seeds/cf_workouts.rb
- git diff --check
- rvm 4.0.5 do bundle exec rubocop db/seeds.rb db/seeds/benchmark_workouts.rb db/seeds/hero_workouts.rb db/seeds/cf_workouts.rb

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -6,8 +6,8 @@
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
 User.find_or_create_by(email: 'admin@wod-tracker.com') do |u|
-  u.first_name = "ADMIN"
-  u.last_name = "ADMIN"
+  u.first_name = 'ADMIN'
+  u.last_name = 'ADMIN'
   u.sex = :male
   u.weight = 0
   u.role = :admin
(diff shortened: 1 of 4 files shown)


Yanchek99 (commented):
Nice refactor — the split is faithful to the original (score types, loads, distances, segments, and dates all map 1:1), and moving from local variables to `seed_movement`/`seed_program` with `.fetch` is a real safety improvement: a mistyped name now raises a clear `KeyError` instead of silently creating a junk record. A few optional, non-blocking notes below.

db/seeds/benchmark_workouts.rb
```diff
@@ -0,0 +1,454 @@
...
+
+# ==============================================================================
+# Elizabeth
+# 21-15-9 reps for time
+# 135-lb. cleans
+# Ring dips
+elizabeth = Workout.find_or_create_by(name: 'Elizabeth') do |workout|
+  workout.interval = '21-15-9'
+  workout.score_type = :time
+  workout.exercises.build(movement: seed_movement('Clean Squat'), position: 1, reps: 1, female_load: 95, male_load: 135, load_unit: :lb)
```
Yanchek99: Now that movement names are explicit literals, this is easier to spot: Elizabeth maps to `'Clean Squat'` rather than `'Clean'` (both exist as movements). This faithfully preserves the old `clean` local, so it's not a regression — but Elizabeth is classically just "cleans," so worth confirming it's intentional. Same applies to Linda at line 231.

db/seeds.rb
```diff
@@ -149,730 +149,38 @@
...
-  workout.exercises.build(movement: run, segment: third_window, position: 1,
-                          reps: 1, distance: 200, distance_unit: :meter)
-  workout.exercises.build(movement: l_pullup, segment: third_window, position: 2, reps: 0)
-
-  fourth_window = workout.segments.build(name: '15:00-20:00', time_seconds: 300, position: 4)
-  workout.exercises.build(movement: run, segment: fourth_window, position: 1,
-                          reps: 1, distance: 200, distance_unit: :meter)
-  workout.exercises.build(movement: deficit_pushup, segment: fourth_window, position: 2,
-                          reps: 0, female_distance: 2, male_distance: 4, distance_unit: :inch)
+def seed_movement(name)
```
Yanchek99: Optional: defining `seed_movement`/`seed_program` at the top level adds private methods to `Object` for the whole process. It's a common and acceptable pattern in seed files, so no change needed — just flagging that wrapping them in a small module/namespace would avoid the global leak if you'd prefer.

db/seeds/cf_workouts.rb
```diff
@@ -0,0 +1,149 @@
...
+segmented = Workout.find_or_create_by!(name: 'CFJ-181202') do |workout|
+  workout.score_type = :time
+  workout.exercises.build(movement: seed_movement('Run'), position: 1, reps: 1, distance: 800, distance_unit: :meter)
+  seg = workout.segments.build(rounds: 10, position: 2)
+  workout.exercises.build(movement: seed_movement('Handstand Push-up'), segment: seg, position: 1, reps: 10)
+  workout.exercises.build(movement: seed_movement('Single-leg Squat (Pistol)'), segment: seg, position: 2, reps: 10)
+  workout.exercises.build(movement: seed_movement('Run'), position: 4, reps: 1, distance: 800, distance_unit: :meter)
+end
+
+seed_program('Crossfit Journal').schedules.find_or_initialize_by(workout: segmented).update(posted_at: '02-12-2018')
```
Yanchek99: Minor consistency nit (pre-existing): CF-260622 uses `Date.new(2026, 6, 22)` while these use ambiguous `'02-12-2018'` strings that rely on Rails parsing them as DD-MM-YYYY. Using `Date.new(...)` everywhere would make the intent unambiguous. Out of scope for this PR if you'd rather keep it a pure move.

PR #1714, opened 2026-07-01
Model prescribed load as a canonical magnitude (#1684)

### LINKED ISSUE #1684: Model prescribed load as a canonical magnitude (lb/kg/pood are display conventions)

Yanchek99:
Closes #1684.

## Problem

CrossFit expresses one prescribed load in whichever unit suits the audience — **lb, kg, or pood** — and these are *display conventions for the same prescription*, not different prescriptions. The old model folded the display unit into identity (`exercises.load_unit` enum), so the content fingerprint hashed `(value, unit)` and the same prescription imported as `95 lb` vs `43 kg` produced two `Workout` rows. That blocks content-based dedup (#1673).

The equivalences are **rounded conventions**, not physics (43 kg is published as 95 lb; the same 65 lb was published as both 29 and 29.5 kg across years), so normalization follows CrossFit's **published table**, not arithmetic.

## Approach (decisions confirmed with the issue owner)

- **One canonical backend unit per dimension**: load → **pounds**, travel distance → **meters**. lb is CrossFit's official unit (the 2019 Open 19.5 note: *"The official weight is in pounds... minimum acceptable weights in kilograms are 43 kg (95 lb)"*); kg/pood are inputs/labels.
- **Canonicalize on write/import** through a source-confirmed table.
- **Display unit is a user preference** (`User#unit_system`: imperial → lb, metric → kg). Stored magnitude and workout identity never depend on it.

## Changes

- **Schema**: drop `load_unit` from `exercises`/`movement_logs`; add `load_bearing` boolean (carries the "bears a load" role, including a find-a-max with no fixed value); add `users.unit_system`. Migration backfills `load_bearing` and **converts existing kg-stored loads to canonical lb**.
- **`LoadEquivalence` / `DistanceEquivalence`** POROs normalize lb/kg/pood → lb and mile/km → meters, plus a documented rounded-conversion fallback.
- **Fingerprint** hashes the canonical pounds value + `load_bearing` instead of `(value, unit)`; the migration recomputes every `content_key`.
- **Forms/controllers**: exercise + log forms drop the unit selector, label loads in the viewer's unit, and normalize entered values to lb on save.
- **Display helpers** render canonical pounds in the viewer's unit (byte-identical output for imperial users).
- **Docs**: source-confirmed table in `cf/docs/load-and-distance-equivalence.md` (backed by L1/L2 guides + games.crossfit.com Open pages in `references.md`); decision entry marked implemented.

## Scope notes

- **Distance is intentionally lighter than load**: travel distances (mile/km) normalize to meters on import; `foot`/`inch` distances and all **height** cases (target/box height) are left unchanged — converting a "10-ft target" would change its meaning. No distance display toggle.
- **Fractional pood** (1.5) must be normalized by the importer before the integer `load` column (covered at the module level in `LoadEquivalenceTest`).
- The live JS summary label during editing shows `lb` for a metric user until save; the server render is correct.

## Verification

- `bundle exec rails test` — **215 runs, 0 failures, 0 errors**. New: `load_equivalence_test`, `distance_equivalence_test`, `workout_load_identity_test` (proves lb/kg/pood → one `content_key`), plus display/user-preference tests.
- RuboCop clean on changed files.
- Migration verified up + rollback via `db:migrate:redo`.
- System tests (Selenium) not run locally — covered by CI `test:all`.


Yanchek99:
Review findings:

- `db/migrate/20260701130000_replace_load_unit_with_load_bearing.rb`: recomputing every `content_key` with `workout.content_fingerprint` can violate the existing unique index when two preexisting workouts only differed by `load_unit` and now normalize to the same fingerprint. That is the duplicate class this PR is meant to handle, so the migration can fail mid-deploy with a unique constraint error. Please use duplicate-aware assignment or clear/merge duplicate keys before writing.

- CI is currently red. The system tests still select the removed `Load unit` field in `test/system/exercise_card_summary_formatting_test.rb` and `test/system/sex_specific_workout_values_test.rb`. GitHub Actions reports 3 errors from those stale selectors.

- `app/javascript/controllers/exercise_card_controller.js`: the live exercise-card summary still defaults load display to `lb` after the unit selector was removed. For a metric user, the form label says `kg`, but clicking “Done” renders the entered value as pounds until the server re-renders. The JS needs the same display unit context as the Slim form.

Issue #1674, opened 2026-06-18
Fetcher: retrieve & extract a CrossFit WOD for a given date


Yanchek99:
Fetch and extract a single day's WOD (part of the WOD-scraper epic).

## Scope
`CrossfitWod::Fetcher` (PORO under `app/services` or `lib`): given a `Date`, GET via `Net::HTTP` (timeouts, User-Agent, redirect following) and return a value object `{ date:, slug:, title:, body_html:, body_text:, description:, scaling:, rest_day?: }`.

**Two page formats must both be supported:**
- **Modern** `https://www.crossfit.com/{YYMMDD}` (e.g. `/250618`): body server-rendered into static HTML inside `<article> ... <span class="_text-block...">`, plus a separate coaching `description`. Exposes `previousUrl`/`nextUrl` for calendar walking.
- **Legacy** (older dates) `https://www.crossfit.com/workout/{YYYY}/{MM}/{DD}`: served directly with a smaller template — body in `<div id="{YYYYMMDD}" class="wod active">`, with scaling options inline.
- `/workout/Y/M/D` 301-redirects to `/{YYMMDD}` for recent dates; choose the URL by date or follow the redirect.

- Detect "Rest Day" bodies.
- Expose `previous`/`next` slugs for the backfill walker.

## Acceptance / Tests
- WebMock tests using saved real fixtures for **both layouts**: modern normal WOD, modern multi-part, modern rest day, legacy WOD with scaling.

PR #1719, opened 2026-07-04
Heuristic parser: WOD prose -> structured Workout (#1675)

### LINKED ISSUE #1675: Heuristic parser: WOD prose → structured Workout
**Core/largest issue.** Convert free-form WOD prose into a structured `Workout` graph (part of the WOD-scraper epic).

## Scope
`CrossfitWod::Parser`: turn `body_text` into an unsaved `Workout` with built exercises, following the `db/seeds.rb` pattern. Extract:
- **Format/score:** "For time" → `score_type: :time`; "AMRAP in N min" → `time` + `score_type: :rep`/`:round`; "EMOM / every minute" → `rounds` + `time`; rep ladders "21-15-9" → `interval`; lifting "5-5-5" / "find a 1RM" → `score_type: :weight`.
- **Lines:** `N <movement> [@ load lb/kg] [over X in/ft/m] [N cal]`; map movement names to the existing catalog in `db/seeds.rb` (singularize/normalize; `find_or_create_by` only when genuinely missing).
- **Scaled loads:** "95/65 lb" → `male_load`/`female_load` + `load_unit`.
- **Multi-part workouts** → `Segment`s (see the 250618 interval example) where the model supports it.

Produce records consistent with `app/models/concerns/workout_scoring.rb` predicates (`amrap?`, `emom?`, `rounds_for_time?`, `interval?`).

## Fallback behavior
- **Raw on the workout:** always set `notes` to the raw body so source text is never lost.
- **Structured-or-review:** if the parser cannot build a valid structured workout (or leaves significant content unparsed), do **not** silently degrade — record the attempt in the review queue (see the WodImport issue) with raw body + reason so we can extend the parser.

## Acceptance / Tests
- A **corpus of real WOD bodies** (saved fixtures spanning formats) asserting parsed structure — this corpus is the battle-test deliverable.

Yanchek99:
## Summary
- Adds `CfWod::Parser`, converting a fetched `CfWod::WodPage`'s `body_text` into an unsaved `Workout`/`Exercise`/`Segment` graph, following the `db/seeds.rb` build pattern.
- Returns a `CfWod::ParseResult` (`workout`, `status`, `reason`, `raw_text`) rather than a bare `Workout` — free-form CrossFit prose is genuinely ambiguous in several well-known ways, so the parser is honest about what it couldn't confidently structure instead of guessing. `status` is `:parsed`, `:partial` (built something, `reason` names the gap), or `:failed` (nothing usable, including rest days — checked before any parsing runs). `workout.notes` always holds the raw body text regardless of status, so source text is never silently lost.
- New collaborators, each narrow and single-purpose: `FormatDetector` (score_type/time/rounds/interval/time_cap_seconds from body-level phrases), `LineClassifier` (classifies each line/paragraph), `BodyBuilder` (walks classified paragraphs), `ExerciseBuilder`/`ExerciseLoadAttacher`/`PenaltySegmentBuilder` (per-exercise construction and scaled-load/penalty-segment handling), `LineParser` (one movement line → structured fields), `MovementMatcher` (normalizes prose and matches/creates against the existing `Movement` catalog per the issue's "only when genuinely missing" instruction).
- Deliberately punted to a flagged `:partial` result rather than guessed: ascending ladders, partner/team (`team_size`), event-triggered-penalty reps semantics, and multi-movement shared trailing-load blocks. Reasoning and known limitations (movement hyphenation conventions, bare-generic-term ambiguity like "squats") are documented in `cf/docs/decisions.md`.

Closes #1675.

## Test plan
- [x] `bundle exec rails test test/services/cf_wod/` — 25 runs, 146 assertions, 0 failures (includes the Fetcher's existing tests)
- [x] `bundle exec rubocop` on all new files — no offenses
- [x] Tests run the real fixtures already committed for the Fetcher (#1674) through the actual pipeline, plus seed-file-derived prose (Cindy, Chelsea, Fran, Open 15.1a) for formats the fixtures don't cover
- [x] Manually verified against the live site (`CfWod::Fetcher.call(Date.current)` piped into `CfWod::Parser.call`) against today's actual WOD, a partner Hero workout — correctly flagged `:partial` with accurate reasons


app/services/cf_wod/workout_format_detector.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    def lifting_result
+      return unless LIFTING_PATTERNS.any? { |pattern| body_text.match?(pattern) }
+
+      seconds = TIME_CAP_PATTERNS.filter_map { |pattern| body_text[pattern, 1] }.first
+      blank_result.with(score_type: :weight, time_cap_seconds: seconds&.to_i&.*(60))
+    end
+
+    def for_time_result
```
Yanchek99, line 99: This only detects `for time` when the cue is on the first line, but current CrossFit pages commonly put a named WOD first and the scoring cue on the next paragraph. I reproduced this with the live July 3, 2026 page: body text starts with `Ned`, then `7 rounds for time of:`, and `CfWod::Parser.call` returns `status=partial`, `score_type=nil`, and `Could not determine workout format from body text`. That leaves the returned `Workout` invalid even though the format is a normal for-time workout.

app/services/cf_wod/movement_matcher.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+    end
+
+    def create_if_plausible(candidate)
+      return nil if candidate.split.size > MAX_NEW_MOVEMENT_WORDS
+
+      Movement.find_or_create_by(name: candidate)
+    end
+
+    def normalize(name)
```
Yanchek99, line 61: Because normalization singularizes before the exact lookup, seeded hyphenated plural movement names can miss their exact catalog entries. For example, `knees-to-elbows` becomes `Knees-to-elbow`, misses the seeded `Knees-to-elbows`, then fuzzy-matches multiple records and returns ambiguous; `toes-to-bars` has the same shape. Known movements like these would be dropped from parsed workouts.

Yanchek99 (commented):
I found two parser issues that are worth fixing before merge. Both are inline on the affected code.

app/services/cf_wod/workout_format_detector.rb
Yanchek99: Fixed in b9e786c. `for_time_result` now checks the whole body (not just the first line) for "for time", and `ROUNDS_FOR_TIME_PATTERN` no longer requires "each" (real prose often omits it, e.g. "7 rounds for time of:"). Verified against your exact repro:

```
status=time score_type set, rounds=7
```

One residual thing I noticed while verifying but didn't fix, since it's outside what you flagged here: the body's first line ("Ned", the Hero WOD's honoree name) still gets treated as a bare unmatched-movement candidate and creates a new `Movement` named "Ned" (single word, under the length guard meant to catch misclassified sentences). I don't have a reliable general way to distinguish a proper name from a legitimate single-word movement (e.g. "Burpee", "Snatch") purely structurally, so I left it rather than add a speculative heuristic. Happy to take a pass at it separately if you'd like — just didn't want to silently expand scope on this review.

app/services/cf_wod/movement_matcher.rb
Yanchek99: Fixed in b9e786c. `MovementMatcher` now tries the as-written form (and a hyphen/space swap) as an exact match on both the as-written and singularized forms before falling back to fuzzy search — so it tries "Knees-to-elbows", "Knees-to-elbow", "Knees to elbows", "Knees to elbow" in that order. Confirmed both your examples now resolve correctly:

- `knees-to-elbows` -> exact match on "Knees-to-elbows" (as-written, no singularization needed)
- `toes-to-bars` -> exact match on "Toes to Bar" (singularized + hyphen-to-space swap)

Added two new test cases covering this (`movement_matcher_test.rb`), plus updated two existing assertions that were incidentally relying on the old behavior creating fresh hyphenated movements instead of finding the pre-existing space-separated fixtures.

Yanchek99:
Closing to restart with a simpler design -- the accumulated class count/indirection here grew out of hand across several review rounds. Re-approaching from scratch in a new branch.

Issue #1679, opened 2026-06-18
[Epic] Scrape CrossFit.com daily WOD into the app


Yanchek99:
## Goal
An ActiveJob fetches CrossFit.com's Workout of the Day and inserts it into WOD Tracker as a structured `Workout` (with `Exercise`/`Movement` records) scheduled onto a dedicated "CrossFit.com" program. This gives a steady stream of real, varied workouts that **battle-tests the data model** (AMRAP, for-time, EMOM, intervals, rep ladders, lifting, scaled male/female loads, rest days, multi-part). Level-up goal: an ad-hoc backfill over a date range.

## Key findings (crossfit.com)
- **Two URL/page formats:**
  - Modern `https://www.crossfit.com/{YYMMDD}` (e.g. `/250618`): workout body server-rendered into static HTML inside `<article> ... <span class="_text-block...">`, plus a separate coaching `description`; exposes `previousUrl`/`nextUrl`.
  - Legacy `https://www.crossfit.com/workout/{YYYY}/{MM}/{DD}` (older dates): smaller template, body in `<div id="{YYYYMMDD}" class="wod active">` with scaling inline. `/workout/Y/M/D` 301-redirects to `/{YYMMDD}` for recent dates.
- Body is **free-form prose** (no structured rep/movement data) — prose→`Exercise` mapping is the hard part.
- "Rest Day" pages must be detected, not parsed.
- Plain HTTP GET + Nokogiri is enough; no headless browser.

## Decisions
- **Heuristic (regex/rule-based) parser** now; raw body always kept in `notes`.
- **SolidQueue** (Rails 8 default) for the recurring daily run + ad-hoc backfill.
- Unparseable WODs go to an **admin-reviewable queue (RailsAdmin)** rather than being dropped.

## Acceptance
Daily WOD auto-imported as a structured workout on the CrossFit.com program; ad-hoc backfill available for a date range; unparseable bodies land in the review queue.

## Sub-issues
- [x] #1672 — Add SolidQueue + scraping/test dependencies
- [x] #1673 — Idempotency: source columns on workouts + CrossFit.com program
- [x] #1674 — Fetcher: retrieve & extract a WOD for a given date
- [x] #1675 — Heuristic parser: WOD prose → structured Workout (core)
- [x] #1676 — Admin review queue for unparseable WODs (WodImport)
- [x] #1677 — Daily scrape job + recurring schedule
[...]

Issue #1676, opened 2026-06-18
Admin review queue for unparseable WODs (WodImport)


Yanchek99:
A reviewable surface for WODs the parser can't structure, so we can extend the parser (part of the WOD-scraper epic).

## Scope
- Migration + model (e.g. `WodImport`): `source`, `source_ref` (`YYMMDD` slug), `wod_date`, `raw_html`/`raw_text`, `status` (enum: `imported`/`partial`/`failed`/`rest_day`), `error_message`, optional `workout_id` reference, timestamps; unique index on `[source, source_ref]`. Keep `db/schema.rb` in sync.
- The Fetcher/Parser/Job write a record per scrape attempt; `failed`/`partial` rows are the actionable queue (replaces "flag in notes").
- **RailsAdmin:** the app auto-discovers AR models and authorizes via CanCanCan (`config/initializers/rails_admin.rb`, `app/models/ability.rb`) — ensure the admin ability permits managing the new model so it appears under `/admin`; optionally configure list/filter on `status`.

## Acceptance
- A WOD the parser can't handle produces a `failed`/`partial` `WodImport` visible in RailsAdmin with its raw text and reason.
- A clean parse produces an `imported` row linked to the workout.
- Re-running an attempt updates (not duplicates) the row.

Yanchek99:
# WodImport Review Queue Implementation Plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

**Goal:** Add a `WodImport` model that records failed/partial WOD scrape-parse attempts, giving admins a RailsAdmin review queue for #1676.

**Architecture:** A single new table (`wod_imports`) and model with two responsibilities: hold the raw text and failure reason for an attempt that needs review, and enforce one row per date (re-running an attempt updates, not duplicates). A curated RailsAdmin list view surfaces it. No orchestration/write path — that's #1677.

**Tech Stack:** Rails 8.1, PostgreSQL, Minitest, RailsAdmin 3.3.0, CanCanCan.

## Global Constraints

- Migration class: `ActiveRecord::Migration[8.1]`.
[...]

PR #1721, opened 2026-07-08
Add CfWod::WorkoutParser: heuristic WOD prose -> structured Workout

### LINKED ISSUE #1675: Heuristic parser: WOD prose → structured Workout

Yanchek99:
## Summary

Implements `CfWod::WorkoutParser` (closes #1675), the core/largest piece of the CrossFit.com WOD-scraper epic (#1679). Turns a `CfWod::WodPage`'s free-form `body_text` into a built (unsaved) `Workout` graph, following the manual-construction idiom already used in `db/seeds/cf_workouts.rb`.

- One orchestrator (`WorkoutParser`) plus six single-purpose collaborators (`WorkoutFormatClassifier`, `PartSplitter`, `ExerciseLineParser`, `MovementLookup`, `PrescriptionClauseParser`, `PrescriptionClauseAssigner`), flat under `app/services/cf_wod/`, matching the existing `Fetcher`/`PageParser` sibling convention.
- Strict movement matching (no `find_or_create_by`), mechanical name normalization, and a fail-closed design: anything the heuristics can't confidently parse raises `CfWod::WorkoutParser::UnparseableError` with a specific reason, rather than guessing — unparseable WODs are meant to land in a future admin review queue (#1676).
- A real-WOD test corpus (8 scenarios) spanning AMRAP, EMOM, rep-ladder, find-a-1RM, sequential and time-windowed multi-part segments, and a multi-implement chipper with compound prescription clauses — plus full unit coverage per collaborator.
- Design doc: `docs/superpowers/specs/2026-07-07-cfwod-workout-parser-design.md`. Implementation plan: `docs/superpowers/plans/2026-07-07-cfwod-workout-parser-implementation.md`.

Built via 16 plan tasks (implementer + independent reviewer per task) plus a final whole-branch review that found and fixed one real gap: sex-asymmetric prescription clause counts could previously raise a raw `TypeError` instead of the intended `UnparseableError`.

## Test plan

- [x] `bin/rails test test/services/cf_wod/` — 52 runs, 201 assertions, 0 failures
- [x] `bundle exec rubocop app/services/cf_wod test/services/cf_wod` — 20 files, 0 offenses
- [x] Existing `test/services/cf_wod/fetcher_test.rb` (unrelated to this change) still passes unmodified


app/services/cf_wod/exercise_line_parser.rb
```diff
@@ -0,0 +1,53 @@
...
+      when BARE_MOVEMENT then { movement_name: clean_name(line), reps: 1 }
+      end
+    end
+
+    private
+
+    attr_reader :line
+
+    def distance_then_movement_attributes
+      value, unit, rest = line.match(DISTANCE_THEN_MOVEMENT).captures
```
Yanchek99: Rest can be confusing here. Maybe call it remaining

app/services/cf_wod/exercise_line_parser.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    def movement_then_distance_attributes
+      movement, value, unit = line.match(MOVEMENT_THEN_DISTANCE).captures
+      { movement_name: clean_name(movement), reps: 1, distance: value.delete(',').to_i,
+        distance_unit: DISTANCE_UNITS.fetch(unit.downcase) }
+    end
+
+    def numbered_reps_attributes
+      reps, rest = line.match(NUMBERED_REPS).captures
```
Yanchek99: Rest can be confusing here. Maybe call it remaining

app/services/cf_wod/movement_lookup.rb
```diff
@@ -0,0 +1,22 @@
+module CfWod
+  class MovementLookup
+    def self.call(name) = new(name).lookup
+
+    def initialize(name)
+      @name = name
+    end
+
+    def lookup
+      Movement.find_by(name: normalized_name)
```
Yanchek99, line 15: Do we need to normalize the name. Cant we just query with a loose match.

app/services/cf_wod/part_splitter.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    def segment_part(name: nil, time_seconds: nil, rounds: nil)
+      { segment: true, name: name, time_seconds: time_seconds, rounds: rounds, lines: [] }
+    end
+
+    def window_seconds(match)
+      to_seconds(match[2]) - to_seconds(match[1])
+    end
+
+    def to_seconds(clock)
```
Yanchek99: This isnt very readable. What is a clock.

app/services/cf_wod/part_splitter.rb
```diff
@@ -0,0 +1,77 @@
...
+    def new_part
+      { segment: false, name: nil, time_seconds: nil, rounds: nil, lines: [] }
+    end
+
+    def segment_part(name: nil, time_seconds: nil, rounds: nil)
+      { segment: true, name: name, time_seconds: time_seconds, rounds: rounds, lines: [] }
+    end
+
+    def window_seconds(match)
+      to_seconds(match[2]) - to_seconds(match[1])
```
Yanchek99: This isnt readable. What is match.

app/services/cf_wod/part_splitter.rb
```diff
@@ -0,0 +1,77 @@
+module CfWod
+  class PartSplitter
+    BOILERPLATE_LINE_PATTERNS = [
```
Yanchek99: Weird name. Maybe Irrelevant_Line_Patterns

app/services/cf_wod/prescription_clause_assigner.rb
```diff
@@ -0,0 +1,99 @@
+module CfWod
+  class PrescriptionClauseAssigner
+    # Movement is genuinely load-bearing here, detected by name/line-text for now -- swap this
+    # for the equipment/load-bearing taxonomy (#1629) once it lands, same stopgap already used
+    # by Movement#supports_implement_count?.
+    BARBELL_FAMILY_KEYWORDS = %w[snatch clean jerk squat deadlift press thruster].freeze
```
Yanchek99: Air squat does not have a barbell

app/services/cf_wod/prescription_clause_assigner.rb
```diff
@@ -0,0 +1,99 @@
...
+    BARBELL_CUE_PATTERN = /overhead|front-rack|back-rack/i
+
+    STOPWORDS = %w[a an the to of and or with for on at in].freeze
+    LOAD_UNITS = %i[lb kg].freeze
```
Yanchek99: Pood might be used too for some older kettlebell workouts.

app/services/cf_wod/prescription_clause_assigner.rb
```diff
@@ -0,0 +1,99 @@
...
+    DISTANCE_UNITS = %i[inch foot meter].freeze
```
Yanchek99, line 18: Mile might be used.

app/services/cf_wod/workout_parser.rb
```diff
@@ -0,0 +1,87 @@
...
+    class UnparseableError < StandardError; end
+
+    def self.call(wod_page) = new(wod_page).parse
+
+    def initialize(wod_page)
+      @wod_page = wod_page
+    end
+
+    def parse
+      header, rest = wod_page.body_text.split("\n", 2)
```
Yanchek99: Rest is odd name here

app/services/cf_wod/workout_parser.rb
```diff
@@ -0,0 +1,87 @@
...
+      build_workout_content(workout, attrs, rest)
+      validate_workout!(workout)
+      workout
+    end
+
+    private
+
+    attr_reader :wod_page
+
+    def build_workout_content(workout, attrs, rest)
```
Yanchek99: Rest is an odd name

docs/superpowers/specs/2026-07-07-cfwod-workout-parser-design.md
```diff
@@ -0,0 +1,189 @@
+# CfWod::WorkoutParser — Heuristic WOD Prose → Structured Workout
```
Yanchek99: I dont think we should commit this. If anything it should be part of the github issue

Yanchek99 (commented):
I think we should adjust the rake task so that it fetches and parses now. Also we should update readme.

app/services/cf_wod/prescription_clause_assigner.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    def barbell_family?(line)
+      name = line[:exercise].movement.name.downcase
+      BARBELL_FAMILY_KEYWORDS.any? { |keyword| name.include?(keyword) } || line[:raw_line].match?(BARBELL_CUE_PATTERN)
+    end
+
+    def apply_value(candidate, female_value, male_value)
+      exercise = candidate[:exercise]
+      if LOAD_UNITS.include?(female_value[:unit])
+        exercise.female_load = female_value[:value]
```
Yanchek99: This path accepts `:kg` in `LOAD_UNITS`, but then writes the raw values straight into `female_load` / `male_load`. Those columns are canonical pounds; `ExercisePrescription` only normalizes kg/pood when `load_unit` is set before validation. For example, a parsed `Female 43 kg / Male 61 kg` prescription will validate and store `[43, 61]` instead of `[95, 135]`, so the imported workout has the wrong load and fingerprint. Either set `exercise.load_unit = female_value[:unit]` before assigning loads, or explicitly convert with `LoadEquivalence.to_lb` here.

app/services/cf_wod/prescription_clause_parser.rb
```diff
@@ -0,0 +1,45 @@
+module CfWod
+  class PrescriptionClauseParser
+    MALE_PREFIX = /\A(?:men|male|♂):?\s*/i
+    FEMALE_PREFIX = /\A(?:women|female|♀):?\s*/i
+    CLAUSE_PATTERN = /\A([\d,]+)[\s-](lb|kg|inch|foot|in|ft)\.?\s*(.*?)\.?\z/i
```
Yanchek99: Since this regex advertises `kg` support, it should handle decimal kg values too. The existing load equivalence table includes CrossFit-published values like `22.5`, `29.5`, and `38.5` kg, but `CLAUSE_PATTERN` only matches integers and `clause_value` currently uses `to_i`. That means common kg prescriptions either fail closed or would be truncated if the regex is loosened without changing the conversion. Consider parsing decimal values for load units and preserving them through assignment/conversion.

Yanchek99 (commented):
Thanks for updating the rake task and README. I found one blocking correctness issue around supported load units: kg clauses are accepted by the parser, but the assigner persists the raw kg numbers as canonical pound values. There is also a related parser gap for decimal kg values, which this app already treats as valid load inputs.

app/services/cf_wod/movement_lookup.rb
Yanchek99: Strict matching was a deliberate call during design, not an oversight — worth restating the reasoning since it's non-obvious:

A loose/fuzzy match risks silently binding to the *wrong* specific movement variant (e.g. matching "Squat" loosely could land on "Air Squat" instead of "Back Squat", or "Push-up" instead of "Deficit Push-up") — that's a wrong-data risk, not just a missed-match risk. The whole parser is designed to fail closed: if it can't confidently resolve something, it raises `UnparseableError` so the WOD lands in the admin review queue (#1676) rather than guessing. A human confirms the real movement (or adds an alias) at that point, and the catalog only grows from confirmed matches — never from `find_or_create_by` or a loose match here.

The mechanical normalization (singularize + title-case, no synonym table) is the same philosophy: cheap wins only, defer anything requiring judgment to the review queue.

Happy to revisit if you want a looser match here — just flagging that it'd be a real behavior change (more binds succeed automatically, but with a higher chance of binding to the wrong specific variant) rather than a pure readability fix.

app/services/cf_wod/prescription_clause_assigner.rb
Yanchek99: Good catch on the real-world usage, but this one's a bigger change than it looks: `Exercise#distance_unit` is a DB-backed enum (`app/models/concerns/exercise_prescription.rb:5`, `enum :distance_unit, { meter: 0, foot: 1, inch: 2 }`) — there's no `mile` value at the schema level at all, for any part of the app, not just this parser. Adding it here would mean a migration + enum change to a model this PR doesn't otherwise touch, not just a parser-side fix.

Given that, I'd rather not fold it into this PR. If you want mile support, I think it's worth its own issue (model change + migration + whatever existing distance-rendering code needs updating for the new unit) rather than a quick addition here — happy to file that if you'd like, or let me know if you'd rather I just do it in this PR.

For now: a WOD prescribing miles would fail closed (`UnparseableError`) rather than silently misparse, consistent with how this parser treats anything it doesn't yet support.

app/services/cf_wod/exercise_line_parser.rb
Yanchek99: Renamed to `remaining` in dade960.

app/services/cf_wod/exercise_line_parser.rb
Yanchek99: Renamed to `remaining` in dade960.

app/services/cf_wod/part_splitter.rb
Yanchek99: Renamed to `clock_time_to_seconds(clock_time)` in dade960.

app/services/cf_wod/part_splitter.rb
Yanchek99: Replaced the bare MatchData param with explicit `start_clock`/`end_clock` locals in `segment_for`/`window_seconds` in dade960 — no more opaque `match`.

app/services/cf_wod/part_splitter.rb
Yanchek99: Renamed to `IRRELEVANT_LINE_PATTERNS` (and `boilerplate_line?` -> `irrelevant_line?`) in dade960.

app/services/cf_wod/prescription_clause_assigner.rb
Yanchek99: Real bug — the substring check on "squat" would've matched "Air Squat" too. Replaced the keyword-substring check with an exact `BARBELL_FAMILY_MOVEMENTS` name allowlist (b208b4d), so only actual barbell squat variants (Back/Front/Overhead Squat, Squat Clean, etc.) match — bodyweight "Air Squat" no longer does. Added a test asserting Air Squat is correctly excluded.

app/services/cf_wod/prescription_clause_assigner.rb
Yanchek99: Good catch — added `:pood` alongside `:lb`/`:kg` in b208b4d. `LoadEquivalence` already had a full pood conversion table, it just wasn't wired up here.

app/services/cf_wod/workout_parser.rb
Yanchek99: Renamed to `body` in dade960 (matches the existing `build_from_body(workout, body)` parameter name one call deeper, so it's consistent too, not just clearer).

app/services/cf_wod/workout_parser.rb
Yanchek99: Same rename, see the reply on line 12.

docs/superpowers/specs/2026-07-07-cfwod-workout-parser-design.md
Yanchek99: Removed from the repo in af45376 and posted as a comment on #1675 instead: https://github.com/Yanchek99/wod-tracker/issues/1675#issuecomment-4915151138

app/services/cf_wod/prescription_clause_assigner.rb
Yanchek99: Confirmed and fixed in b208b4d. Verified the exact failure mode you described directly: a test asserting `Movement: Power Snatch, ♀43kg/♂61kg` was storing `[43, 61]` before the fix (should be `[95, 135]`).

Went with explicit `LoadEquivalence.to_lb(value, unit)` conversion in `apply_value` (for every unit in `LOAD_UNITS`, now `lb`/`kg`/`pood`) rather than the `exercise.load_unit=`/`before_validation` mechanism — turned out `load_unit=` can't handle decimal kg values reliably either (see the reply on the decimal-kg comment), since `female_load`/`male_load` are integer columns that truncate a Float on assignment, before the callback ever runs. Converting explicitly before assignment sidesteps that entirely and matches how `db/seeds.rb` already assigns canonical lb values directly.

Added three new tests: whole-kg conversion (43/61 -> 95/135), decimal-kg conversion (22.5/38.5 -> 50/85), and pood conversion.

app/services/cf_wod/prescription_clause_parser.rb
Yanchek99: Fixed in b208b4d. Extended `CLAUSE_PATTERN` to accept an optional decimal portion (`[\d,]+(?:\.\d+)?`) and `clause_value` now parses to a Float when the raw text has a decimal point, Integer otherwise — this matters because `LoadEquivalence::KG_TO_LB` keys on the *exact* magnitude (`22.5`, `29.5`, `38.5` are separate hash keys from `22`, `29`, `38`), so truncating to an integer before conversion would miss those lookups and silently fall through to the generic 5lb-rounding fallback instead of the published value. Added a test asserting `22.5`/`38.5` kg round-trip to the correct `50`/`85` lb via the real published table, not the fallback.

Yanchek99:
Addressed both rounds of review:

**Blocking correctness issues (both fixed in b208b4d, with new tests):**
- kg/pood loads were accepted but never converted to canonical pounds (`♀43kg/♂61kg` stored `[43, 61]` instead of `[95, 135]`) — now converts through `LoadEquivalence.to_lb` for every load unit.
- Decimal kg values (`22.5`, `29.5`, `38.5` — real published CrossFit conversions) were truncated to integers before conversion, missing `LoadEquivalence`'s exact-magnitude lookup table — now preserved as floats through parsing.
- Also added `:pood` support alongside `lb`/`kg` while in there (per your other comment).

**Also fixed from the first round of inline comments:**
- `BARBELL_FAMILY_KEYWORDS`'s substring match on "squat" incorrectly matched the bodyweight "Air Squat" — replaced with an exact barbell-movement-name allowlist.
- Renamed the various `rest`/`clock`/`match`/`BOILERPLATE_LINE_PATTERNS` identifiers per your comments.
- Removed the design spec from the repo; posted it to #1675 instead: https://github.com/Yanchek99/wod-tracker/issues/1675#issuecomment-4915151138
- Replied inline on the two open questions (loose movement matching, mile distance support) rather than just implementing — wanted to flag the tradeoffs before changing behavior.

All existing + new tests pass (63 runs across `test/services/cf_wod/`), rubocop clean.

app/services/cf_wod/prescription_clause_assigner.rb
Yanchek99: Good point — no schema change needed after all: mile doesn't need to be its own canonical `distance_unit`, it can just convert to meters at parse time, same as `LoadEquivalence` already collapses kg/pood into canonical pounds rather than storing them as their own unit. Implemented in b94fe82 (1 mile -> 1600m, matching CrossFit's own gym-floor rounding rather than the precise 1609.34m), with tests for both line-grammar orders and a decimal-mile case (1.5 mile -> 2400m).

PR #1739, opened 2026-07-10
Parse open-ended ascending ladders from CrossFit WOD prose

### LINKED ISSUE #1729: CfWod::WorkoutParser: no heuristic for open-ended ascending-rep ladders ("Etc.")
## Problem

\`CfWod::WorkoutParser\` has no heuristic for parsing an open-ended ascending-rep ladder from prose. CrossFit.com signals this pattern by writing out the first few rungs explicitly and ending with the literal word \`"Etc."\`, e.g. today's WOD (260710):

\`\`\`
Complete as many reps as possible in 10 minutes of:
3 burpee box jump-overs
3 deadlifts
6 burpee box jump-overs
6 deadlifts
9 burpee box jump-overs
9 deadlifts
Etc.
\`\`\`

Parsing this currently raises \`CfWod::WorkoutParser::UnparseableError: unrecognized exercise line: "Etc."\` — \`"Etc."\` isn't a recognized exercise line, and there's no logic anywhere in the parser to detect the repeating-couplet-with-increasing-reps pattern and collapse it into the app's existing ascending-ladder model (\`Workout#ladder_step\`, \`#ascending_ladder?\`, \`ladder_step_every\` on \`Exercise\` — currently only used by hand-authored Open-workout seed data, per \`db/seeds/open_workouts.rb\`).

Found via #1727 while running the new \`ScrapeCfWodJob\`/\`cf_wod:scrape\` rake task against the live site. A narrower, unrelated header-recognition gap in the same WOD ("as many reps as possible" AMRAP phrasing) was fixed separately in #1728 — that fix does NOT resolve this issue; the body still fails on \`"Etc."\` after the header parses correctly.

## Why this is more than a quick fix

This isn't a small regex/pattern addition like most of this parser's existing format handlers. It requires:
- Detecting the repeating-group-with-increasing-reps structure across multiple exercise lines (not just classifying a single header line)
- Inferring \`ladder_step\` (and which exercises the ladder applies to per round) from the explicit rungs shown
- Deciding how many explicit rungs are "enough" to confirm the pattern before trusting \`"Etc."\` as the terminator, and failing closed (this parser's established philosophy — see \`cf/docs/decisions.md\` and existing corpus tests) if the rungs shown don't cleanly imply a single, unambiguous step size
- New corpus test coverage for this shape, plus probably case-by-case review of how it interacts with the app's rest of the existing part/segment splitting logic

This should go through the same brainstorm → design → plan cycle the original \`CfWod::WorkoutParser\` used (issue #1675), not a quick patch.

Yanchek99:
## Summary

Fixes #1729 by parsing CrossFit.com open-ended ascending-rep ladders that list initial rungs and end with `Etc.`.

- Adds `CfWod::AscendingLadderInferer` to validate repeated ascending rungs and infer `Workout#ladder_step`.
- Integrates the inference into `CfWod::WorkoutParser` before exercise construction, preserving existing prescription assignment for collapsed first-rung exercises.
- Adds unit, parser, and corpus coverage for the CF-260710 shape plus fail-closed ambiguity cases.
- Adds design and implementation plan docs under `docs/superpowers/`.

## Root Cause

`WorkoutParser` previously treated every body content line as an exercise line. A terminal `Etc.` line reached `ExerciseLineParser`, produced no attributes, and raised `UnparseableError` instead of recognizing the cross-line ascending ladder structure.

## Validation

- `rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel`  
  Result: `144 files inspected, no offenses detected` after merging current `origin/master`.
- `rvm 4.0.5@wod-tracker do bundle exec rails test:all`  
  Result before merging current `origin/master`: `364 runs, 1042 assertions, 0 failures, 0 errors, 0 skips`.
- Post-merge `rails test:all` hit one transient PostgreSQL deadlock in `test/system/workout_drag_ordering_test.rb`; rerunning that file passed: `6 runs, 20 assertions, 0 failures, 0 errors, 0 skips`.
- Focused parser/rake suites also passed during implementation: `99 runs, 323 assertions, 0 failures, 0 errors, 0 skips`.

## Notes

Final code review found no blocking issues. A non-blocking follow-up suggestion was to add explicit tests for periodless `Etc` and exactly two-rung valid ladders.


app/services/cf_wod/ascending_ladder_part_normalizer.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      [part.merge(lines: ladder[:lines].pluck(:raw_line))]
+    end
+
+    private
+
+    attr_reader :workout, :parts
+
+    def ladder_candidate?
+      parts.one? && !part[:segment] && part[:lines].last&.match?(ETC_LINE)
```
Yanchek99, line 26: **Important — missing test coverage for part/segment interaction, which issue #1729 explicitly asked for.**

`ladder_candidate?` requires `parts.one? && !part[:segment]`, so any multi-part body (e.g. a `"Then, N rounds of:"` split) or a segment-wrapped body (e.g. `"0:00-10:00:"`) that ends in `Etc.` is silently skipped and falls through to the existing `UnparseableError`.

I confirmed by hand that this is currently correct — but there's no committed test asserting it. Issue #1729 asked for coverage of "interaction with existing part/segment splitting logic" in addition to the base corpus shape; that half wasn't delivered. Since the behavior is currently safe, this is a regression-safety gap rather than a live bug, but a future refactor of `PartSplitter` or this predicate could silently break it with nothing to catch it.

Suggest: add 1-2 tests asserting `UnparseableError` for a segment-wrapped ladder and a multi-part (`"Then,"`) body ending in `Etc.`.

app/services/cf_wod/ascending_ladder_inferer.rb
```diff
@@ -0,0 +1,77 @@
...
+          next_line.dig(:attrs, :reps) - previous_line.dig(:attrs, :reps)
+        end
+      end
+      raise_ambiguous unless deltas.uniq.one? && deltas.first&.positive?
+
+      deltas.first
+    end
+
+    def raise_ambiguous
+      raise WorkoutParser::UnparseableError, 'ambiguous ascending ladder ending in Etc.'
```
Yanchek99: **Minor — single generic error message across five distinct failure modes.**

Every validation failure (too few rungs, reordered movements, non-uniform step, non-increasing reps, malformed reps) raises the identical `'ambiguous ascending ladder ending in Etc.'`. Per `ScrapeCfWodJob`/`WodImport`, this message becomes `WodImport#error_message`, surfaced in the admin review queue (#1726) specifically so a human can diagnose failed scrapes — a generic message makes that triage harder than it needs to be.

Not worth blocking on, but consider including which check failed, e.g. `"ambiguous ascending ladder ending in Etc.: rep step is not constant across rungs"`.

Yanchek99 (commented):
Reviewed in an isolated worktree (`test/services/cf_wod/` — 96 runs, 314 assertions, 0 failures; RuboCop clean). Core ladder-inference algorithm is correct and genuinely fail-closed — traced several adversarial inputs (reordered movements, non-uniform deltas, descending reps, single-movement ladders) and all correctly raise rather than silently guessing. Integration point is minimal and non-candidate bodies are unaffected. Two inline notes below; neither blocks merge, but the first is worth a quick fast-follow.

Issue #1717, opened 2026-07-03
Simplify leading-work-metric selection in measurable helpers


Yanchek99:
## Context

Flagged during review of #1716.

`leading_work_metric_stands_alone?` (app/helpers/measurable_leading_work_helper.rb, since renamed to `prescribed_work_metric_stands_alone?` in app/helpers/measurable_prescribed_work_helper.rb) and its mirror in app/javascript/controllers/exercise_card_controller.js decide whether a distance/calorie metric can stand alone as the "leading" prescription text. The logic is a growing OR-chain of exceptions:

```ruby
def prescribed_work_metric_stands_alone?(candidate, metrics)
  candidate && metrics.all? do |metric|
    !visible_metric?(metric) ||
      metric == candidate ||
      structural_single_rep_metric?(metric) ||
      load_detail_for_prescribed_work?(candidate, metric, metrics)
  end
end
```

`load_detail_for_prescribed_work?` is itself a 3-part AND mixing "is candidate distance/calorie", "is this metric a load", and "does candidate have a scalar value OR is there a structural rep elsewhere". Every new exception (like the load carve-out added in #1716) has to be added to this formula *and* re-derived by hand in the parallel JS implementation.

## Concrete symptom (fixed in #1716)

An earlier version of #1716 had a real bug caused by this duplication: the `ft` compact-formatting exception was applied to the non-sex-specific path but not the sex-specific path, in both Ruby and JS, because the formatting was written twice instead of shared. That specific bug was fixed by extracting a shared `prescribed_work_metric_unit_msg` / `prescribedWorkUnitText` helper used by both paths — a good example of the kind of drift this duplication invites, and worth watching for as more exceptions get added.

## Suggested direction

Consider consolidating "pick the leading metric, then decide what else may accompany it" into a single small model (e.g. a `PrescribedWork` value object owning its own formatting and load-compatibility) instead of continuing to add boolean exceptions to `prescribed_work_metric_stands_alone?` / the JS equivalent. Not urgent, but worth doing before a third exception gets bolted on.

Yanchek99:
# Leading Prescription Refactor Design

## Context

Issue #1717 asks for simplifying leading-work-metric selection in measurable helpers. The current logic for distance/calorie leading text lives in `MeasurablePrescribedWorkHelper` and is mirrored by hand in `exercise_card_controller.js`. It works, but the selection rule has become a growing chain of exceptions:

- a distance or calorie metric may lead the movement text
[...]

PR #1742, opened 2026-07-11
Simplify leading prescription selection

Yanchek99:
## Summary

- Add `Measurable::LeadingPrescription` to own non-duration leading prescription selection.
- Route Ruby exercise summaries and additional metrics through the new object while preserving existing ordering and literal single-rep behavior from `master`.
- Mirror the same selection model in the exercise-card JavaScript preview.
- Cover max reps, max calories, distance/calorie leading text, structural single reps, load details, foot formatting, additional metric ordering, and exact local-preview max-calorie output.

## Why

Issue #1717 called out that leading-work-metric selection had become a duplicated Ruby/JavaScript chain of boolean exceptions. This keeps the behavior stable while making the selection model explicit and easier to compare across server rendering and local preview rendering.

## Verification

- `rvm 4.0.5@wod-tracker do bundle exec rails test test/helpers/measurable_helper_test.rb test/helpers/measurable_prescribed_work_helper_test.rb`
  - `31 runs, 41 assertions, 0 failures, 0 errors, 0 skips`
- `rvm 4.0.5@wod-tracker do bundle exec rails test test/system/exercise_card_summary_formatting_test.rb`
  - `6 runs, 7 assertions, 0 failures, 0 errors, 0 skips`
- `yarn build`
- `git diff --check`


Yanchek99:
## Code Review

**Reviewed range:** `c5ea155`..`fd6dba3`

### Strengths
- `Measurable::LeadingPrescription` genuinely consolidates selection + formatting per issue #1717's suggested direction, not just a rename.
- Old `measurable_prescribed_work_helper.rb` cleanly removed with zero dangling references (verified by grep across the PR head).
- The specific past bug class (sex-specific vs. non-sex-specific `ft` formatting drift, from #1716) is structurally guarded via shared unit-text methods in both Ruby and JS, not just avoided by convention.
- Test claims verified independently in a throwaway worktree at the exact head SHA: `test/helpers/measurable_helper_test.rb` + `test/helpers/measurable_prescribed_work_helper_test.rb` — 31 runs, 41 assertions, 0 failures; `test/system/exercise_card_summary_formatting_test.rb` — 6 runs, 7 assertions, 0 failures. Matches the PR description.

### Issues

#### Important (Should Fix)
- **`app/helpers/measurable_helper.rb:6`** — `measurable_movement_msg` newly trips `Metrics/AbcSize` (21.4/20) and `Metrics/CyclomaticComplexity` (8/7). Confirmed base commit (`c5ea155`) has zero RuboCop offenses in this file; PR head (`fd6dba3`) has two. The PR's own verification steps didn't run RuboCop, so this will surface for the first time in CI's reviewdog gate. Fix: extract the `if leading&.metric` block (lines 12–17) into a small private method, or restructure with early returns.

#### Minor (Nice to Have)
- **`app/helpers/measurable_helper.rb:93`** — `movement_name_for_summary_metric` has two unreachable `metric.rep?` branches; the only call site already filters reps out before calling it. Dead conditionals make the method look more general than it is.
- **`app/javascript/controllers/exercise_card_controller.js:211-213`** — `additionalMetrics` builds a second `LeadingPrescription` instance just to fall back to `leading.metric`, but that fallback is unreachable given current call sites. Low cost today but reads as meaningful defensive logic when it isn't.

### Recommendations
The Ruby/JS duplication isn't eliminated, just made isomorphic (matching class/method names/order across `Measurable::LeadingPrescription` and the JS mirror). That's a reasonable, pragmatic reading of #1717's "single small model" suggestion and lowers divergence risk, but a third exception would still need to be hand-added to both implementations, and nothing (shared fixture, cross-language golden test) enforces they stay in sync. Worth confirming this partial resolution is acceptable, since #1717 named cross-language drift as the root cause of a past bug.

### Assessment

**Ready to merge?** With fixes

**Reasoning:** Core implementation is solid — independently verified both test suites pass at the exact head commit, and the merge-commit-plus-followup-fix sequence composes correctly with the concurrently-merged "literal single rep" feature. The one blocking item is a genuine, newly-introduced RuboCop violation that CI will flag; it's a small, mechanical fix.

---
🤖 Posted via requested code review

PR #1745, opened 2026-07-11
Match repeated Open workouts titled "Open Workout <n.n>" to their catalog name

Yanchek99:
## Summary
- `cf_wod:scrape[2026-07-11]` failed with `Failed: unrecognized format header: "Open Workout 20.5"` — today's WOD is a rerun of Open 20.5.
- crossfit.com titles Open-workout reruns "Open Workout 20.5" while the catalog stores them under their plain name "Open 20.5", so `find_named_workout` missed the match and fell through to prose parsing, which then failed on the `"For time, partitioned any way:"` header format used only by these reruns.
- `find_named_workout` now also tries stripping the "Workout " crossfit.com inserts before giving up, returning the existing cataloged workout directly instead of re-deriving (or duplicating) it from fragile prose.

## Test plan
- [x] `bundle exec rails test test/services/cf_wod/workout_parser_test.rb test/services/cf_wod/workout_parser_corpus_test.rb`
- [x] `bundle exec rubocop app/services/cf_wod/workout_parser.rb test/services/cf_wod/workout_parser_test.rb`
- [x] Verified against the live page: `bin/rails "cf_wod:fetch[2026-07-11]"` now returns the existing "Open 20.5" catalog record instead of raising

### DIFF
--- a/app/services/cf_wod/workout_parser.rb
+++ b/app/services/cf_wod/workout_parser.rb
@@ -2,11 +2,6 @@ module CfWod
   class WorkoutParser
     class UnparseableError < StandardError; end
 
-    # crossfit.com titles an Open-workout rerun "Open Workout <n.n>" (e.g. "Open Workout 20.5"),
-    # while the catalog stores it under its plain name (e.g. "Open 20.5") -- the same convention
-    # used for every other Open workout.
-    OPEN_RERUN_TITLE = /\AOpen Workout (.+)\z/i
-
     def self.call(wod_page) = new(wod_page).parse
 
     def initialize(wod_page)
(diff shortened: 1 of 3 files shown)


app/services/cf_wod/workout_parser.rb
```diff
@@ -34,7 +34,12 @@ def find_named_workout(lines)
       title = lines.first.to_s.strip
       return nil if title.blank? || recognized_header?(title)
 
-      Workout.find_by(name: title)
+      Workout.find_by(name: title) || find_open_rerun_workout(title)
+    end
+
+    # crossfit.com titles an Open-workout rerun "Open Workout 20.5"; the catalog stores it as "Open 20.5".
+    def find_open_rerun_workout(title)
+      Workout.find_by(name: title.sub(/\bWorkout /i, '')) if title.match?(/\AOpen\b/i)
```
Yanchek99: Why can't we just search for a loose match on the number portion for open workouts.  So if the title contains dd.d then search *dd.d*

Yanchek99: Good call — updated to match loosely on the number (e.g. `20.5`) instead of stripping the word "Workout". One wrinkle: the catalog has trailing-letter variants like "Open 15.1a" alongside "Open 15.1", so a plain `%15.1%` substring search would ambiguously match both. Anchored the LIKE at the end (`%20.5` instead of `%20.5%`) so it can't cross that boundary — see 9a5b6da.

PR #1758, opened 2026-07-14
Backfill implicit segments for top-level exercises

### LINKED ISSUE #1752: Segment unification Task 1: Backfill implicit segments for top-level exercises
## Goal

Data-only migration: backfill an implicit `Segment` for every existing top-level `Exercise` (`segment_id: nil`), so every exercise ends up attached to a segment. This is step 1 of the design in #1750 — it prepares the data so later tasks can stop allowing `segment_id: nil` without breaking existing workouts. No schema constraint or model/validation change happens in this task; that's deliberately deferred (see Background).

**Files:**
- Create: `db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb`
- Create: `test/migrations/backfill_segments_for_top_level_exercises_test.rb`

**Interfaces:**
- Produces: every `Exercise` row has a non-null `segment_id` after this migration runs (enforced by the migration's own data, not yet by a DB constraint or model validation).
- Consumes: none (first task in the epic, no dependency).

## Background: why the NOT NULL constraint is NOT in this task

The obvious design is "backfill, then add `NOT NULL`" as one migration (that's how the existing `RestructureSegmentsAsStructuralBlocks` migration adds a column, backfills, and drops the old column in one `up`). That pattern is unsafe here specifically: until Task 3 removes the model-level `optional: true` and the app stops offering "add exercise directly to a workout" (Task 5's form change), the running app can still create a new `Exercise` with `segment_id: nil` at any time. Adding a DB-level `NOT NULL` in this task would make the *current, unchanged* form start raising `ActiveRecord::NotNullViolation` on ordinary save the moment this migration deploys — before any of the code that stops creating such rows has landed.

So the constraint is added later, in Task 4 (dropping dead columns), by which point Task 3 has already made `belongs_to :segment` required at the model level and Task 5 has removed the form's top-level-exercise path. This task only makes the *data* already conform; enforcement comes after the *code* conforms.

## Migration design

Reconstructs each affected workout's current `ordered_parts` sequence (top-level exercises + segments, sorted by `position` — the same merge `Workout#ordered_parts` does today), groups consecutive top-level exercises into "runs", and wraps each run in one new `Segment`.

Key simplification: a run's new segment can reuse the **first exercise's existing `position`** value as its own `position`. Positions were already validated unique across top-level exercises *and* segments together in the current model (that's the entire purpose of today's `position_unique_within_workout_parts`), so no renumbering or collision-checking is needed — the value is guaranteed free among that workout's segment positions already.

**Scheme assignment for a new segment:** a workout's own `rounds`/`time`/`interval` describe *the run of top-level exercises*, never any of its segments (segments already carry their own independent scheme). Checked against real seed data (`db/seeds/hero_workouts.rb`, `open_workouts.rb`): every workout that mixes top-level exercises with real segments has at most one exercise run, except when the workout's own scheme is the `rounds: 1`-with-no-time-no-interval sentinel (meaning "no real top-level scheme" — see "Brenton" vs "Alec" below). So:
- If the workout has a real scheme (`rounds > 1`, or `time` present, or `interval` present) — copy it onto the run's new segment (`rounds` direct; `time_seconds = time * 60`, since `Workout#time` is minutes and `Segment#time_seconds` is seconds; `interval_scheme = interval` direct, same `"21-15-9"` string format both sides use).
- Otherwise (the sentinel case) — the new segment is unschemed (`rounds`/`time_seconds`/`interval_scheme` all nil), matching how `Workout#rounds_for_time?`/`for_time_objective` already treat `rounds: 1` as "just say For Time," not "1 round of."
- If a workout somehow has *more than one* exercise run **and** a real scheme, the migration raises rather than guessing — this hasn't occurred in current seed/fixture data, but real user-entered workouts could differ, and silently mis-assigning a scheme to the wrong run would corrupt content silently.

Worked examples from real seed data:
- **Fran** (`interval: '21-15-9'`, no segments): one run = the whole workout → new segment gets `interval_scheme: '21-15-9'`.
- **Brenton** (`rounds: 5`, one segment named "After every 5 broad jumps"): one exercise run (bear crawl, broad jump) → new segment gets `rounds: 5`; the existing penalty segment is untouched.
- **Alec** (`rounds: 1` sentinel, two segments): one exercise run (run/sandbag/run, sandwiched between the two real segments) → new segment is unschemed, just groups those three exercises; the two existing segments are untouched.

```ruby
# db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
class BackfillSegmentsForTopLevelExercises < ActiveRecord::Migration[8.1]
  class MigrationWorkout < ActiveRecord::Base
    self.table_name = 'workouts'
  end

  class MigrationSegment < ActiveRecord::Base
    self.table_name = 'segments'
    self.record_timestamps = false
  end

  class MigrationExercise < ActiveRecord::Base
    self.table_name = 'exercises'
    self.record_timestamps = false
  end

  def up
    workout_ids = MigrationExercise.where(segment_id: nil).distinct.pluck(:workout_id)

    say_with_time "Backfilling implicit segments for #{workout_ids.size} workouts" do
      workout_ids.each { |workout_id| backfill_workout(workout_id) }
    end
  end

  def down
    say 'Data-only migration -- implicit segments created by #up are not restorable to top-level exercises. No-op.'
  end

  private

  def backfill_workout(workout_id)
    workout = MigrationWorkout.find(workout_id)
    runs = group_runs(ordered_parts(workout_id))
    exercise_runs = runs.select { |run| run[:type] == :exercises }

    if exercise_runs.size > 1 && workout_scheme(workout)
      raise "Workout #{workout_id} has #{exercise_runs.size} top-level-exercise runs and a non-sentinel " \
            "scheme (rounds=#{workout.rounds}, time=#{workout.time}, interval=#{workout.interval}) -- " \
            'ambiguous which run the scheme belongs to. Needs manual review before this migration can proceed.'
    end

    exercise_runs.each { |run| wrap_run(workout, run) }
  end

  # Mirrors today's Workout#ordered_parts: top-level exercises and segments together, sorted by position.
  def ordered_parts(workout_id)
    exercises = MigrationExercise.where(workout_id: workout_id, segment_id: nil).order(:position)
                                  .map { |record| { type: :exercise, record: } }
    segments = MigrationSegment.where(workout_id: workout_id).order(:position)
                                .map { |record| { type: :segment, record: } }
    (exercises + segments).sort_by { |part| part[:record].position }
  end

  # Collapses consecutive top-level exercises into one :exercises run; segments pass through untouched.
  def group_runs(parts)
    parts.chunk { |part| part[:type] }.map do |type, group|
      { type: type == :exercise ? :exercises : :segment, records: group.map { |part| part[:record] } }
    end
  end

  # nil when the workout's own scheme is the "no real top-level scheme" sentinel (rounds <= 1, no time/interval).
  def workout_scheme(workout)
    return nil if workout.rounds.to_i <= 1 && workout.time.blank? && workout.interval.blank?

    {
      rounds: (workout.rounds if workout.rounds.to_i > 1),
      time_seconds: (workout.time * 60 if workout.time.present?),
      interval_scheme: workout.interval
    }
  end

  def wrap_run(workout, run)
    scheme = workout_scheme(workout) || {}
    segment = MigrationSegment.create!(workout_id: workout.id, position: run[:records].first.position, **scheme)

    run[:records].each_with_index do |exercise, index|
      exercise.update!(segment_id: segment.id, position: index + 1)
    end
  end
end
```

## Steps

- [ ] **Step 1: Write the failing test**

```ruby
# test/migrations/backfill_segments_for_top_level_exercises_test.rb
require 'test_helper'
require Rails.root.join('db/migrate/20260713120000_backfill_segments_for_top_level_exercises').to_s

class BackfillSegmentsForTopLevelExercisesTest < ActiveSupport::TestCase
  test 'wraps a whole flat workout in one segment carrying its interval scheme' do
    workout = Workout.create!(name: 'Fran', score_type: :time, interval: '21-15-9')
    workout.exercises.create!(movement: movements(:thruster), position: 1, reps: 1, load: 95)
    workout.exercises.create!(movement: movements(:pullup), position: 2, reps: 1)

    BackfillSegmentsForTopLevelExercises.new.up

    workout.reload
    assert_equal 1, workout.segments.count
    segment = workout.segments.first
    assert_equal '21-15-9', segment.interval_scheme
    assert_nil segment.rounds
    assert_equal [1, 2], segment.exercises.order(:position).pluck(:position)
    assert_equal 0, Exercise.where(workout_id: workout.id, segment_id: nil).count
  end

  test 'wraps a real top-level round count onto its own segment, leaving an existing segment untouched' do
    workout = Workout.create!(name: 'Brenton', score_type: :time, rounds: 5)
    workout.exercises.create!(movement: movements(:run), position: 1, reps: 1, distance: 30, distance_unit: :meter)
    penalty = workout.segments.create!(position: 2, name: 'After every 5 reps')
    penalty.exercises.create!(movement: movements(:pullup), position: 1, reps: 3)

    BackfillSegmentsForTopLevelExercises.new.up

    workout.reload
    assert_equal 2, workout.segments.count
    new_segment = workout.segments.find { |segment| segment.id != penalty.id }
    assert_equal 5, new_segment.rounds
    assert_equal 1, new_segment.position
    assert_equal 'After every 5 reps', penalty.reload.name
    assert_equal 2, penalty.position
  end

  test 'leaves a flat stretch between two segments unschemed' do
    workout = Workout.create!(name: 'Alec', score_type: :time, rounds: 1)
    triplet = workout.segments.create!(position: 1, rounds: 3)
    triplet.exercises.create!(movement: movements(:pushup), position: 1, reps: 9)
    workout.exercises.create!(movement: movements(:run), position: 2, reps: 1, distance: 1000, distance_unit: :meter)
    finisher = workout.segments.create!(position: 5, rounds: 5)
    finisher.exercises.create!(movement: movements(:pullup), position: 1, reps: 10)

    BackfillSegmentsForTopLevelExercises.new.up

    workout.reload
    assert_equal 3, workout.segments.count
    middle = workout.segments.find { |segment| segment.position == 2 }
    assert_nil middle.rounds
    assert_nil middle.time_seconds
    assert_equal [movements(:run)], middle.exercises.map(&:movement)
  end

  test 'raises when multiple exercise runs coexist with a real top-level scheme' do
    workout = Workout.create!(name: 'Ambiguous', score_type: :time, rounds: 3)
    workout.exercises.create!(movement: movements(:run), position: 1, reps: 1, distance: 400, distance_unit: :meter)
    workout.segments.create!(position: 2, rounds: 2)
    workout.exercises.create!(movement: movements(:pullup), position: 3, reps: 5)

    assert_raises(RuntimeError) { BackfillSegmentsForTopLevelExercises.new.up }
  end
end
```

- [ ] **Step 2: Run the tests to verify they fail**

```bash
bin/rails test test/migrations/backfill_segments_for_top_level_exercises_test.rb
```

Expected: `LoadError` / file-not-found on the `require` line (the migration file doesn't exist yet).

- [ ] **Step 3: Write the migration** (full code above)

- [ ] **Step 4: Run the tests to verify they pass**

```bash
bin/rails test test/migrations/backfill_segments_for_top_level_exercises_test.rb
```

- [ ] **Step 5: Run the migration against the dev/seed database and spot-check**

```bash
bin/rails db:migrate
bin/rails runner '
  puts "Remaining top-level exercises: #{Exercise.where(segment_id: nil).count}"
  puts Workout.find_by(name: "Alec").segments.order(:position).map { |s| [s.position, s.rounds, s.time_seconds, s.exercises.count] }.inspect
'
```

Expected: `0` remaining top-level exercises; Alec shows 3 segments (rounds 3 / unschemed / rounds 5).

- [ ] **Step 6: Commit**

```bash
git add db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb test/migrations/backfill_segments_for_top_level_exercises_test.rb
git commit -m "Backfill implicit segments for top-level exercises"
```

## Verification

- `bin/rails test test/migrations/backfill_segments_for_top_level_exercises_test.rb` passes
- `bin/rails db:migrate` runs clean against the seeded dev database with zero remaining `Exercise.where(segment_id: nil)` rows
- `bundle exec rubocop db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb` passes

---
Part of #1751

Yanchek99:
## Summary
- add a data-only migration that wraps existing top-level exercise runs in implicit segments
- copy workout-level schemes onto generated segments, including time-to-seconds conversion
- refresh affected workout content keys after restructuring
- add focused migration tests for the requested examples and regressions

## Verification
- rvm 4.0.5@wod-tracker do bundle exec rails test test/migrations/backfill_segments_for_top_level_exercises_test.rb
- rvm 4.0.5@wod-tracker do bundle exec rubocop db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
- git diff --check

Closes #1752

### DIFF
--- a/app/models/concerns/implicit_workout_part.rb
+++ b/app/models/concerns/implicit_workout_part.rb
@@ -4,7 +4,7 @@ module ImplicitWorkoutPart
   # Backfilled wrappers preserve the old workout-part shape: the workout owns the
   # scheme/objective, and this segment only groups the exercises that used to be top-level.
   def implicit_workout_part?
-    blank_wrapper? && workout.present? && segment_scheme == workout_part_scheme
+    persisted? && blank_wrapper? && workout.present? && segment_scheme == workout_part_scheme
   end
 
   private
@@ -22,14 +22,9 @@ module ImplicitWorkoutPart
   end
 
   def workout_part_scheme
-    return {} if unschemed_weight_workout?
-    return nil if unschemed_workout?
+    return {} if unschemed_workout?
 
     workout_real_scheme
-  end
-
-  def unschemed_weight_workout?
-    workout.score_measurement == 'weight' && unschemed_workout?
   end
 
   def unschemed_workout?
--- a/app/models/concerns/implicit_workout_part.rb
+++ b/app/models/concerns/implicit_workout_part.rb
@@ -1,0 +1,54 @@
+module ImplicitWorkoutPart
+  extend ActiveSupport::Concern
+
+  # Backfilled wrappers preserve the old workout-part shape: the workout owns the
+  # scheme/objective, and this segment only groups the exercises that used to be top-level.
+  def implicit_workout_part?
+    blank_wrapper? && workout.present? && segment_scheme == workout_part_scheme
+  end
+
+  private
+
+  def blank_wrapper?
+    name.blank? && rest_seconds.blank? && notes.blank?
+  end
+
+  def segment_scheme
+    {
+      rounds: rounds,
+      time_seconds: time_seconds,
+      interval_scheme: interval_scheme.presence
+    }.compact
+  end
+
+  def workout_part_scheme
+    return {} if unschemed_weight_workout?
+    return nil if unschemed_workout?
+
+    workout_real_scheme
+  end
+
+  def unschemed_weight_workout?
+    workout.score_measurement == 'weight' && unschemed_workout?
+  end
+
+  def unschemed_workout?
+    workout.rounds.to_i <= 1 && workout.time.blank? && workout.interval.blank?
+  end
+
+  def workout_real_scheme
+    {
+      rounds: workout_rounds,
+      time_seconds: workout_time_seconds,
+      interval_scheme: workout.interval.presence
+    }.compact
+  end
+
+  def workout_rounds
+    workout.rounds if workout.rounds.to_i > 1
+  end
+
+  def workout_time_seconds
+    workout.time * 60 if workout.time.present?
+  end
+end
--- a/db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
+++ b/db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
@@ -1,0 +1,80 @@
+class BackfillSegmentsForTopLevelExercises < ActiveRecord::Migration[8.1]
+  class MigrationWorkout < ActiveRecord::Base
+    self.table_name = 'workouts'
+  end
+
+  class MigrationSegment < ActiveRecord::Base
+    self.table_name = 'segments'
+    self.record_timestamps = false
+  end
+
+  class MigrationExercise < ActiveRecord::Base
+    self.table_name = 'exercises'
+    self.record_timestamps = false
+  end
+
+  def up
+    workout_ids = MigrationExercise.where(segment_id: nil).distinct.pluck(:workout_id)
+
+    say_with_time "Backfilling implicit segments for #{workout_ids.size} workouts" do
+      workout_ids.each { |workout_id| backfill_workout(workout_id) }
+    end
+  end
+
+  def down
+    say 'Data-only migration -- implicit segments created by #up are not restorable to top-level exercises. No-op.'
+  end
+
+  private
+
+  def backfill_workout(workout_id)
+    workout = MigrationWorkout.find(workout_id)
+    exercise_runs = exercise_runs(workout_id)
+    scheme = workout_scheme(workout)
+    reject_ambiguous_runs!(workout, exercise_runs, scheme)
+    exercise_runs.each { |run| wrap_run(workout, run, scheme) }
+  end
+
+  def ordered_parts(workout_id)
+    (MigrationExercise.where(workout_id:, segment_id: nil).to_a + MigrationSegment.where(workout_id:).to_a)
+      .sort_by { |part| [part.position, part.id] }
+  end
+
+  def exercise_runs(workout_id)
+    ordered_parts(workout_id)
+      .chunk_while { |left, right| left.is_a?(MigrationExercise) && right.is_a?(MigrationExercise) }
+      .select { |run| run.first.is_a?(MigrationExercise) }
+  end
+
+  def reject_ambiguous_runs!(workout, exercise_runs, scheme)
+    return unless exercise_runs.size > 1 && scheme
+
+    raise "Workout #{workout.id} has #{exercise_runs.size} top-level exercise runs with rounds=#{workout.rounds.inspect}, " \
+          "time=#{workout.time.inspect}, interval=#{workout.interval.inspect}"
+  end
+
+  def workout_scheme(workout)
+    return if workout.rounds.to_i <= 1 && workout.time.blank? && workout.interval.blank?
+
+    {
+      rounds: (workout.rounds if workout.rounds.to_i > 1),
+      time_seconds: (workout.time * 60 if workout.time.present?),
+      interval_scheme: workout.interval
+    }.compact
+  end
+
+  def wrap_run(workout, run, scheme)
+    segment = MigrationSegment.create!(
+      {
+        workout_id: workout.id,
+        position: run.first.position,
+        created_at: Time.current,
+        updated_at: Time.current
+      }.merge(scheme || {})
+    )
+
+    run.each_with_index do |exercise, index|
+      exercise.update!(segment_id: segment.id, position: index + 1)
+    end
+  end
+end
--- a/db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
+++ b/db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
@@ -1,4 +1,6 @@
 class BackfillSegmentsForTopLevelExercises < ActiveRecord::Migration[8.1]
+  disable_ddl_transaction!
+
   class MigrationWorkout < ActiveRecord::Base
     self.table_name = 'workouts'
   end
@@ -14,7 +16,7 @@ class BackfillSegmentsForTopLevelExercises < ActiveRecord::Migration[8.1]
   end
 
   def up
-    workout_ids = MigrationExercise.where(segment_id: nil).distinct.pluck(:workout_id)
+    workout_ids = MigrationExercise.where(segment_id: nil).distinct.pluck(:workout_id).compact
 
     say_with_time "Backfilling implicit segments for #{workout_ids.size} workouts" do
       workout_ids.each { |workout_id| backfill_workout(workout_id) }
@@ -33,7 +35,7 @@ class BackfillSegmentsForTopLevelExercises < ActiveRecord::Migration[8.1]
     scheme = workout_scheme(workout)
     reject_ambiguous_runs!(workout, exercise_runs, scheme)
     exercise_runs.each { |run| wrap_run(workout, run, scheme) }
-    Workout.find(workout_id).refresh_content_key!
+    refresh_and_absorb_duplicate!(workout_id)
   end
 
   def ordered_parts(workout_id)
@@ -60,8 +62,15 @@ class BackfillSegmentsForTopLevelExercises < ActiveRecord::Migration[8.1]
     {
       rounds: (workout.rounds if workout.rounds.to_i > 1),
       time_seconds: (workout.time * 60 if workout.time.present?),
-      interval_scheme: workout.interval
+      interval_scheme: workout.interval.presence
     }.compact
+  end
+
+  def refresh_and_absorb_duplicate!(workout_id)
+    Workout.find(workout_id).tap do |workout|
+      workout.save!(validate: false)
+      workout.absorb_duplicate!
+    end
   end
 
   def wrap_run(workout, run, scheme)
--- a/db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
+++ b/db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
@@ -33,6 +33,7 @@ class BackfillSegmentsForTopLevelExercises < ActiveRecord::Migration[8.1]
     scheme = workout_scheme(workout)
     reject_ambiguous_runs!(workout, exercise_runs, scheme)
     exercise_runs.each { |run| wrap_run(workout, run, scheme) }
+    Workout.find(workout_id).refresh_content_key!
   end
 
   def ordered_parts(workout_id)
--- a/app/helpers/segments_helper.rb
+++ b/app/helpers/segments_helper.rb
@@ -5,6 +5,10 @@ module SegmentsHelper
     "#{msg}#{segment_prescription(segment)}"
   end
 
+  def segment_objective?(segment)
+    !segment.implicit_workout_part?
+  end
+
   private
 
   def named_max_reps_segment?(segment)
--- a/app/models/concerns/workout_scoring.rb
+++ b/app/models/concerns/workout_scoring.rb
@@ -35,7 +35,10 @@ module WorkoutScoring
   end
 
   def top_level_exercises
-    exercises.reject(&:segment_id)
+    direct_exercises = exercises.reject(&:segment_id)
+    return direct_exercises if direct_exercises.any?
+
+    segments.select(&:implicit_workout_part?).flat_map(&:exercises)
   end
 
   def amrap_score_components
--- a/app/models/segment.rb
+++ b/app/models/segment.rb
@@ -1,4 +1,5 @@
 class Segment < ApplicationRecord
+  include ImplicitWorkoutPart
   include RefreshesWorkoutContentKey
 
   belongs_to :workout
--- a/app/views/workouts/_segment_fields.html.slim
+++ b/app/views/workouts/_segment_fields.html.slim
@@ -1,5 +1,6 @@
 - expanded = f.object.new_record?
-- summary = f.object.persisted? ? segment_objective(f.object) : 'New segment'
+- implicit_wrapper = f.object.implicit_workout_part?
+- summary = f.object.persisted? ? (implicit_wrapper ? 'Exercises' : segment_objective(f.object)) : 'New segment'
 - exercise_summaries = f.object.exercises.reject(&:marked_for_destruction?).sort_by { |exercise| exercise.position || 0 }.map { |exercise| measurable_message(exercise) }
 
 .nested-fields.card.mb-3.workout-part data-controller="segment-card" data-segment-card-expanded-value=expanded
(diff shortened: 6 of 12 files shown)


db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  private
+
+  def backfill_workout(workout_id)
+    workout = MigrationWorkout.find(workout_id)
+    exercise_runs = exercise_runs(workout_id)
+    scheme = workout_scheme(workout)
+    reject_ambiguous_runs!(workout, exercise_runs, scheme)
+    exercise_runs.each { |run| wrap_run(workout, run, scheme) }
```
Yanchek99, line 37: Critical: this unconditionally wraps every top-level exercise run into a segment, which breaks Workout#top_level_exercises (app/models/concerns/workout_scoring.rb:37-38, `exercises.reject(&:segment_id)`) for every workout touched.

That method backs exercises_for_log_recording, top_level_max_finding?, amrap_score_components, and fixed_amrap_reps_per_round. Concretely, the back_squat_5x5 fixture (rounds: 5, score_type: weight, one top-level exercise) is set_based_lifting? today, and exercises_for_log_recording returns the exercise repeated 5 times via top_level_exercises. After this migration wraps it into an implicit segment, top_level_exercises becomes [], so exercises_for_log_recording returns [] and lifting_score/log.rb's build_movement_logs silently stop recording sets for that workout — the same pattern breaks top_level_max_finding?-based max-load workouts and rep-scored AMRAPs (amrap_score_components, fixed_amrap_reps_per_round).

None of this migration's own tests exercise WorkoutScoring, so this regression would ship silently. Worth confirming whether WorkoutScoring is meant to be updated to look at segment exercises before this backfill runs in production.

db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
```diff
@@ -0,0 +1,81 @@
...
+  def exercise_runs(workout_id)
+    ordered_parts(workout_id)
+      .chunk_while { |left, right| left.is_a?(MigrationExercise) && right.is_a?(MigrationExercise) }
+      .select { |run| run.first.is_a?(MigrationExercise) }
+  end
+
+  def reject_ambiguous_runs!(workout, exercise_runs, scheme)
+    return unless exercise_runs.size > 1 && scheme
+
+    raise "Workout #{workout.id} has #{exercise_runs.size} top-level exercise runs with rounds=#{workout.rounds.inspect}, " \
```
Yanchek99, line 55: This migration never calls disable_ddl_transaction!, so (per Rails default behavior on Postgres, which supports transactional DDL) the whole `up` method runs inside one implicit transaction. If any single workout hits this ambiguous-run case, the raise here unwinds all segments already created for every other workout processed earlier in the same run — not just this one. For a large backfill, one bad workout means the entire migration does zero net work, and the only test coverage (backfill_segments_for_top_level_exercises_test.rb, 'raises for multiple exercise runs') only asserts the exception is raised, not what happens to progress on other workouts in the same run.

Worth considering whether ambiguous workouts should be logged/collected and skipped so one bad row doesn't roll back an otherwise-clean backfill of thousands of workouts, requiring a manual data fix + redeploy cycle per ambiguous workout found.

db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  private
+
+  def backfill_workout(workout_id)
+    workout = MigrationWorkout.find(workout_id)
+    exercise_runs = exercise_runs(workout_id)
+    scheme = workout_scheme(workout)
+    reject_ambiguous_runs!(workout, exercise_runs, scheme)
+    exercise_runs.each { |run| wrap_run(workout, run, scheme) }
+    Workout.find(workout_id).refresh_content_key!
```
Yanchek99: This recomputes the content key via the real Workout#refresh_content_key!, but doesn't follow through with absorb_duplicate! the way the codebase's own precedent for this exact class of change does. See db/migrate/20260701130000_drop_load_unit.rb's `up`, which explicitly does `workout.save!(validate: false); workout.absorb_duplicate!` after a bulk restructuring that could cause fingerprint collisions, with the comment: "Recompute and merge through the model ... instead of writing the raw key, so a genuine duplicate is folded into its canonical workout rather than tripping the unique index."

Here, wrapping top-level exercises into a segment changes canonical_content's shape (a bare `{exercise: {...}}` entry becomes `{segment: {..., exercises: [...]}}`), so a workout that previously fingerprinted uniquely can now collide with another workout that already had an equivalent explicit segment. assignable_content_key (workout_fingerprint.rb:50-56) silently sets content_key: nil in that case rather than raising — the migration has no way to detect or merge that duplicate, so it would go silently un-deduped.

db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  def reject_ambiguous_runs!(workout, exercise_runs, scheme)
+    return unless exercise_runs.size > 1 && scheme
+
+    raise "Workout #{workout.id} has #{exercise_runs.size} top-level exercise runs with rounds=#{workout.rounds.inspect}, " \
+          "time=#{workout.time.inspect}, interval=#{workout.interval.inspect}"
+  end
+
+  def workout_scheme(workout)
+    return if workout.rounds.to_i <= 1 && workout.time.blank? && workout.interval.blank?
```
Yanchek99, line 60: Note that workout_scheme returning nil here does not stop the workout's exercise runs from being wrapped — backfill_workout (line 35) calls wrap_run(workout, run, scheme) unconditionally regardless of whether scheme is nil. So every workout with any top-level exercise gets an implicit segment, even ones with no rounds/time/interval at all. That's probably intentional per the PR description ("wraps existing top-level exercise runs in implicit segments"), but it does mean a plain "for time" workout with a flat exercise list now renders a segment with no name — see the top-level review comment for the resulting SegmentsHelper fallback text ("Segment:") this introduces on the show page.

db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    {
+      rounds: (workout.rounds if workout.rounds.to_i > 1),
+      time_seconds: (workout.time * 60 if workout.time.present?),
+      interval_scheme: workout.interval
```
Yanchek99: rounds and time_seconds are nil-guarded (`if workout.rounds.to_i > 1` / `if workout.time.present?`), but `interval_scheme: workout.interval` is not — it's only filtered by .compact afterward, which strips nil but not blank strings. If workout.interval is ever "" rather than nil (the workouts.interval column has no NOT NULL/normalization), the segment ends up with interval_scheme: "" instead of nil. That's meaningfully different from nil for Segment#timed_rounds? (app/models/segment.rb:36, which checks `interval_scheme.nil?`, not `.blank?`) and for the content fingerprint — a hand-authored segment with interval_scheme: nil and a migrated one with interval_scheme: "" covering the same prescription would no longer fingerprint identically.

db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb
```diff
@@ -0,0 +1,81 @@
...
+    self.record_timestamps = false
+  end
+
+  class MigrationExercise < ActiveRecord::Base
+    self.table_name = 'exercises'
+    self.record_timestamps = false
+  end
+
+  def up
+    workout_ids = MigrationExercise.where(segment_id: nil).distinct.pluck(:workout_id)
```
Yanchek99: Minor/edge case: exercises.workout_id has no NOT NULL constraint in the schema (only the app-level belongs_to :workout enforces it on normal writes). If any exercise row ever ended up with workout_id: nil and segment_id: nil (e.g. via a console fix or a bug predating this validation), this pluck would include nil in workout_ids, and backfill_workout(nil) would raise ActiveRecord::RecordNotFound from MigrationWorkout.find(nil) — aborting the whole migration run per the transaction concern noted on line 53. Might be worth a .compact here for defense in depth; low priority given it depends on pre-existing bad data.

Yanchek99 (commented):
Automated review of the segment-backfill migration. The migration itself is well-tested for its own scope (segment creation, content-key refresh, ambiguous-run detection), but tracing its effects through the rest of the app surfaced a few issues, most importantly a scoring regression that isn't covered by the migration's own tests. Two of the findings below touch files outside this diff (they can't be attached as inline comments on unchanged lines), so I'm listing them here:

**app/models/concerns/workout_scoring.rb:37-38 — top_level_exercises breaks for every backfilled workout with a scheme.** top_level_exercises is defined as exercises.reject(&:segment_id). This migration moves every top-level exercise into a segment (see inline comment on line 35), so after the backfill runs, top_level_exercises returns [] for any workout it touches. That method backs exercises_for_log_recording (used by Log#build_movement_logs and lifting_score), top_level_max_finding?, amrap_score_components, and fixed_amrap_reps_per_round. Concretely: the back_squat_5x5 fixture (rounds: 5, score_type: weight, one top-level exercise) is set_based_lifting? today, and exercises_for_log_recording returns rounds.times.flat_map { top_level_exercises } = the exercise repeated 5 times. After this migration wraps it into an implicit segment, top_level_exercises is [], so exercises_for_log_recording returns [], and lifting_score (log_scoring.rb:85 calls workout.amrap_score_components, and log.rb:16 calls exercises_for_log_recording) silently stops recording/scoring sets for that workout. None of the migration's own tests exercise WorkoutScoring, so this regression wouldn't be caught by the migration test suite.

**View double-rendering / _form.html.slim field duplication.** Because wrap_run runs unconditionally for every exercise run regardless of whether workout_scheme returned a scheme (see line 35/58 comments), two visible regressions follow: (1) app/views/workouts/show.html.slim:18 renders workout_objective(@workout) from the workout's untouched rounds/time/interval, and then line 27 renders segment_objective(part, ...) from the segment's copied scheme — for a workout like back_squat_5x5 this now shows the round count twice ("5 sets for load" followed by "5 rounds of"). (2) For a plain workout with no scheme at all (rounds/time/interval all blank), the migration still creates an unnamed segment; SegmentsHelper#segment_prescription (app/helpers/segments_helper.rb:18) falls through to "#{segment.name.presence || 'Segment'}:", so every such backfilled workout gains a literal "Segment:" bullet above its exercise list where there used to be a flat list. Separately, app/views/workouts/_form.html.slim:8-10 (workout-level rounds/time/interval inputs) and _segment_fields.html.slim (rendered per-segment via line 41-44) now both expose editable copies of the same scheme for a backfilled workout — editing one without the other will desync the workout's own fields from the segment's, and both are folded into the content fingerprint (workout_fingerprint.rb:82-90).

Both of the above stem from the same root cause flagged inline on db/migrate/.../20260713120000_backfill_segments_for_top_level_exercises.rb line 35.

app/models/concerns/implicit_workout_part.rb
```diff
@@ -0,0 +1,54 @@
...
+  def segment_scheme
+    {
+      rounds: rounds,
+      time_seconds: time_seconds,
+      interval_scheme: interval_scheme.presence
+    }.compact
+  end
+
+  def workout_part_scheme
+    return {} if unschemed_weight_workout?
```
Yanchek99: This only special-cases `unschemed_weight_workout?` — for any other unschemed score type (e.g. a plain `:time` or `:rep` chipper with no rounds/time/interval), `workout_part_scheme` falls through to `return nil if unschemed_workout?` on the next line, which never equals `segment_scheme`'s `{}` for a blank auto-generated wrapper segment. So `implicit_workout_part?` is `false` for that shape, and (per `segments_helper.rb`'s `segment_objective?` / the `_form.html.slim` `implicit_wrapper` check) the migration's blank wrapper segment still renders its own "Segment:" line on the show page and a mostly-empty segment editor card in the edit form for ordinary flat "for time"/rep-scored workouts.

I confirmed this doesn't affect scoring — `rep_scored_amrap?`/`max_finding?` both require conditions (`time` present, `score_measurement == 'weight'`) that are mutually exclusive with `unschemed_workout?`, so `top_level_exercises` returning `[]` for this shape is harmless for `WorkoutScoring`. It's purely a view regression, but likely a common one given how many workouts are plain flat chippers.

The new test `segments_helper_test.rb`'s "keeps objectives for explicit unnamed unschemed segments" locks in the current behavior (unnamed+unschemed non-weight segments are NOT treated as implicit) — worth confirming whether that's an intentional trade-off (can't safely distinguish a real blank segment from an implicit wrapper without a scheme to compare) or whether `unschemed_weight_workout?` should generalize to all score types, e.g. `workout.rounds.to_i <= 1 ... ? {} : ...` regardless of `score_measurement`, if there's no other reliable signal that separates a genuinely user-authored blank segment from this migration's wrapper.

Yanchek99 (commented):
Re-reviewed after `2310308` ("Address segment backfill review feedback"). Ran the migration/model/helper tests plus RuboCop on the changed files — all green.

5 of the 6 original findings are fully resolved, verified with the new tests added in this commit:

- **Scoring regression** — fixed via the new `ImplicitWorkoutPart` concern; `Workout#top_level_exercises` now falls back to the wrapping segment's exercises when it matches the workout's own scheme. `test/models/implicit_workout_part_test.rb` confirms lifting score, max-finding, and fixed-AMRAP all still compute correctly through the wrapper.
- **Transaction rollback risk** — `disable_ddl_transaction!` added, so one ambiguous workout no longer rolls back previously-processed workouts.
- **Silent dedup gap** — now calls `workout.save!(validate: false); workout.absorb_duplicate!`, matching the `drop_load_unit` precedent; the new "absorbs a duplicate" test confirms the merge actually happens.
- **Blank interval string** — `workout.interval.presence` guards it now; confirmed nil via the new "normalizes blank interval strings" test.
- **Nil workout_id edge case** — `.compact` added after the pluck; confirmed via the new "ignores malformed top-level exercises" test.

One is only partially resolved — see the inline comment on `implicit_workout_part.rb:25`. I reproduced it directly (script below) rather than just reading the code:

```
workout = Workout.create!(name: 'Plain Chipper', score_type: :time)  # no rounds/time/interval
# ... two top-level exercises, then run the migration ...
segment.implicit_workout_part?  # => false
segment_objective?(segment)     # => true
segment_objective(segment)      # => "Segment:"
```

I confirmed this no longer affects any scoring/logging path (`rep_scored_amrap?`/`max_finding?`/`exercises_for_log_recording` all check conditions — amrap? needing time present, weight-scoring — that are mutually exclusive with `unschemed_workout?`, so `top_level_exercises` returning `[]` here is harmless). It's a view-only regression, but a fairly common one: it'll surface on every plain "for time"/rep-scored chipper with no rounds/time/interval that gets backfilled, showing a bare "Segment:" bullet above the exercise list where a flat list used to render, plus an empty segment editor card in the edit form.

Issue #1757, opened 2026-07-13
Segment unification Task 6: Fixtures, controller tests, and full regression pass


Yanchek99:
## Goal

Rewrite the workout/segment/exercise fixtures to the segment-only shape, fix the three controller tests that still submit top-level `exercises_attributes` or read `Workout#rounds`, and do a full-suite regression pass — including the system tests, which need hands-on review against the real rendered form rather than blind rewriting (see the flagged list below).

Depends on: Task 3 (#1754) and Task 5 (#1756) merged.

**Files:**
- Modify: `test/fixtures/workouts.yml`
- Modify: `test/fixtures/segments.yml`
- Modify: `test/fixtures/exercises.yml`
- Modify: `test/controllers/workout_merge_test.rb`
- Modify: `test/controllers/workout_position_updates_test.rb`
- Modify: `test/controllers/workouts_controller_test.rb`
- Review/modify: 9 system tests under `test/system/` that exercise the workout builder form (list below)

## Fixture rewrite

Every workout fixture drops `rounds`/`time`/`interval` (columns still exist until Task 4, but nothing should populate them anymore); every exercise fixture moves from `workout:` to `segment:`. Each exercise fixture also **keeps** its `workout:` key for now, redundantly — the column isn't dropped until Task 4, and Rails fixtures bypass model validations on load, so there's no harm in leaving it set. **Task 4 should delete the `workout:` key from every entry in `exercises.yml` once `exercises.workout_id` is actually dropped** (a two-line addition to that task, not repeated here).

```yaml
# test/fixtures/workouts.yml (full file)
fran:
  name: Fran
  score_type: time

murph:
  name: Murph
  score_type: time

segmented:
  name: Segmented
  score_type: time

amrap_couplet:
  name: AMRAP Couplet
  score_type: rep

amrap_mixed:
  name: AMRAP Mixed
  score_type: rep

amrap_unknown_distance:
  name: AMRAP Unknown Distance
  score_type: rep

back_squat_5x5:
  name: Back Squat 5x5
  score_type: weight

segmented_total_reps:
  name: Segmented Total Reps
  score_type: rep
```

```yaml
# test/fixtures/segments.yml (full file)
fran_segment:
  workout: fran
  interval_scheme: 21-15-9
  position: 1

murph_segment:
  workout: murph
  position: 1

segmented_before:
  workout: segmented
  position: 1

test:
  workout: segmented
  rounds: 10
  position: 2

segmented_after:
[...]

Issue #1756, opened 2026-07-13
Segment unification Task 5: Segment-only workout show page, builder form, and controller params


Yanchek99:
## Goal

Update the workout show page, the workout builder form, the schedule/program listing partials, and controller params so nothing reads `Workout#ordered_parts` or a top-level `Exercise`, and the form stops writing to `Workout#rounds`/`#time`/`#interval` (still present as columns until Task 4, but no longer the target of any input once this task lands). Also resolves the rendering question left open in #1750 for an unschemed, unnamed segment used as a flat stretch mid-workout.

Depends on: Task 3 (#1754) merged — needs `Workout#governing_segment`, `Workout#segments` as the one ordered sequence, `Segment#schemed?`.

**Files:**
- Modify: `app/helpers/workouts_helper.rb`
- Modify: `app/helpers/segments_helper.rb`
- Modify: `app/views/workouts/show.html.slim`
- Modify: `app/views/workouts/_form.html.slim`
- Modify: `app/views/workouts/_exercise_fields.html.slim`
- Modify: `app/views/workouts/_segment_fields.html.slim`
- Modify: `app/views/exercises/_exercises.html.slim`
- Modify: `app/controllers/workouts_controller.rb`
- Investigate/modify: `app/javascript/controllers/ladder_fields_controller.js` (see JS section below — flagged, not fully specified)

**Interfaces:**
- Consumes: `Workout#governing_segment`, `Segment#schemed?` (Task 3).
- Produces: `SegmentsHelper#segment_objective` can now return `nil` (an unschemed, unnamed segment renders no header line — see below); callers must handle that.
- Produces: `WorkoutsController#new` builds a workout with one default segment already attached, so the form always has somewhere to put "the first block" without an extra click for the common single-segment case.

## `WorkoutsHelper` rewrite

Every method that read `workout.rounds`/`workout.time`/`workout.interval` directly now reads `workout.governing_segment`'s equivalent field. `Workout#time` was minutes; `Segment#time_seconds` is seconds, so display math needs `/ 60` where the old code used the value directly.

```ruby
# app/helpers/workouts_helper.rb
module WorkoutsHelper
  def workout_objective(workout)
    return ascending_ladder_objective(workout) if workout.ascending_ladder?
    return max_finding_objective(workout) if workout.max_finding?
    return "#{pluralize workout.governing_segment.rounds, 'set'} for load" if workout.set_based_lifting?
[...]

Issue #1755, opened 2026-07-13
Segment unification Task 4: Drop workout scheme columns and exercises.workout_id


Yanchek99:
## Goal

Final cleanup migration: add the DB-level `NOT NULL` on `exercises.segment_id` (safe now that nothing creates a segment-less exercise), then drop `workouts.rounds`/`time`/`interval` and `exercises.workout_id` entirely, plus the small model code that only existed to keep `workout_id` in sync.

Depends on: Task 3 (#1754) **and** Task 5 (views/forms) merged and deployed — the workout form (`_form.html.slim`) still binds `f.input :rounds`/`:time`/`:interval` directly to `Workout` attributes today, so those columns can't be dropped until Task 5 removes those inputs. (Task 3 alone is not enough to safely run this migration — the tracking issue's dependency list undersells this; Task 4 needs both 3 and 5.)

**Files:**
- Create: `db/migrate/20260713150000_drop_workout_scheme_and_exercise_workout_id.rb`
- Modify: `app/models/exercise.rb`
- Modify: `app/models/concerns/exercise_position_validation.rb`
- Modify: `app/models/log.rb`
- Modify: `app/controllers/workouts_controller.rb` (minor: eager-loading query shape)

**Interfaces:**
- Removes: `Workout#rounds`/`#time`/`#interval` (columns gone — any remaining caller is a bug from a task that didn't fully land; this migration is the final proof no such caller exists).
- Removes: `Exercise#workout`/`#workout_id` (association and column gone).

## Migration

```ruby
# db/migrate/20260713150000_drop_workout_scheme_and_exercise_workout_id.rb
class DropWorkoutSchemeAndExerciseWorkoutId < ActiveRecord::Migration[8.1]
  def up
    change_column_null :exercises, :segment_id, false

    remove_index :exercises, name: 'index_exercises_on_position_and_workout_id'
    remove_index :exercises, name: 'index_exercises_on_position_and_segment_id'
    add_index :exercises, %i[position segment_id], unique: true, name: 'index_exercises_on_position_and_segment_id'

    remove_foreign_key :exercises, :workouts
    remove_column :exercises, :workout_id

    remove_column :workouts, :rounds
    remove_column :workouts, :time
    remove_column :workouts, :interval
  end

  def down
    add_column :workouts, :interval, :string
    add_column :workouts, :time, :integer
    add_column :workouts, :rounds, :integer

    add_column :exercises, :workout_id, :bigint
[...]

Issue #1754, opened 2026-07-13
Segment unification Task 3: Delegate workout scheme predicates to a governing segment


Yanchek99:
## Goal

Rewire the model/concern layer to stop reading `Workout`'s own `rounds`/`time`/`interval` and stop treating top-level `Exercise` as a first-class case: `Workout` delegates its scheme-family predicates to a single "governing segment," `Exercise#segment` becomes required, and the fingerprint/position-validation/scoring concerns drop their top-level-exercise branches. This is the core code change from #1750.

Depends on: Task 1 (#1752) merged and deployed — the data must already be backfilled before the model stops allowing `segment_id: nil`.

**Files:**
- Modify: `app/models/workout.rb`
- Modify: `app/models/segment.rb`
- Modify: `app/models/exercise.rb`
- Modify: `app/models/concerns/workout_scoring.rb`
- Modify: `app/models/concerns/workout_fingerprint.rb`
- Modify: `app/models/concerns/workout_position_reservation.rb`
- Modify: `app/models/concerns/exercise_position_validation.rb`

**Interfaces:**
- Produces: `Workout#governing_segment` (new, public) — the one segment that determines the workout's overall scheme, or `nil` when no single segment dominates (a genuine multi-part chipper like Alec). Task 5's `WorkoutsHelper`/`SegmentsHelper` will call this.
- Produces: `Segment#schemed?` (new) — `true` if the segment has any of `rounds`/`time_seconds`/`interval_scheme` set.
- Removes: `Workout#ordered_parts` (Task 5 must stop calling it — `workout.segments` is the one sequence now).
- Removes: `Workout has_many :exercises` as a direct, writable, FK-based association — becomes `has_many :exercises, through: :segments` (read-only convenience).

## Background: what "governing segment" resolves and why

Every workout's overall shape (is it AMRAP, EMOM, rounds-for-time, an interval ladder, or a plain flat "for time" pass) used to come from `Workout`'s own `rounds`/`time`/`interval` columns, which only had a real, non-sentinel value when the workout had no segments. Now that every workout is built from segments, that same overall shape has to come from *one* segment — but the design's own worked examples (Task 1, Task 2) show three genuinely different shapes:

- **Fran**: one segment, carries the real scheme (`interval_scheme: '21-15-9'`).
- **Murph**: one segment, unschemed (a flat, once-through list of movements) — was `workout.rounds: 1` (the sentinel), now nothing on the segment at all.
[...]

PR #1760, opened 2026-07-15
Complete segment unification: governing segment, form/views, fixtures (Tasks 3, 5, 6)

Yanchek99:
## Summary

Originally scoped to #1754 (Task 3 of the "Segment unification" epic, #1751), this branch grew to cover the rest of the epic after code review and CI investigation surfaced that leaving it at Task 3 alone would ship a red-CI, partially-broken intermediate state. It now completes Tasks 3, 5, and 6 of #1751 end-to-end:

**Task 3 (#1754) — core model rewiring:** every `Exercise` now requires a `Segment`; `Workout` delegates its scheme-family predicates (`amrap?`, `rounds_for_time?`, `emom?`, etc.) to a single "governing segment" via the new `Workout#governing_segment`/`Segment#schemed?`. Also fixes two bugs found during code review: the CF WOD parser wasn't carrying the header-classified scheme onto the segment it builds, and `CollectionProxy#one?/#many?/#count` silently return wrong answers for an unsaved workout with only just-built segments (exactly the parser's own case).

**Task 5 (#1756) — segment-only views/controller:** the workout show page, builder form, and controller params no longer read the deleted `Workout#ordered_parts` or a top-level `Exercise`; `WorkoutsController#new` builds a default segment so the common single-segment case needs no extra click. Also fixes a real bug the task review caught: the ladder-fields JS's initial hidden-state precompute used `governing_segment` instead of "first segment" (what the JS itself targets), causing a visible flash on some real multi-segment workouts.

**Task 6 (#1757) — fixtures, controller tests, full regression:** `test/fixtures/{workouts,segments,exercises}.yml` rewritten to the segment-only shape; three controller tests updated to submit `segments_attributes` instead of top-level `exercises_attributes`/`workout.rounds`; a real N+1-style query regression in `schedules_controller.rb` (exposed by fixtures finally having real segments) fixed by consolidating on one eager-load path; an orphaned test file for a method deleted back in Task 3 removed (explicit authorization obtained first — it also tests a still-live method by a similar name, so this wasn't done unilaterally); and all system tests that drive the real rendered builder form updated against the live page (not blind rewriting) — both the 9 files the issue flagged and 2 more discovered by running the full system suite.

## Verification

- `bin/rails test:all` (matches CI exactly): **421 runs, 0 failures, 0 errors, 0 skips**
- `bundle exec rubocop --parallel`: **150 files inspected, 0 offenses**
- CI on this PR: Build, Analyze (actions/js/ruby), CodeQL, rubocop — all green
- Manual browser verification (Selenium/headless Chrome + direct interactive testing): Fran/Alec/Brenton show pages render correctly (schemed segment, unschemed mid-workout segment with no header, named penalty segment); `/workouts/new` form has no top-level scheme fields and only "Add Segment" at the top level; ladder-fields AMRAP-shape toggle works correctly for create and edit, single- and multi-segment; full create+edit round trip via the form succeeds end-to-end with zero console errors

This is the epic's (#1751) final gate — the full suite is green end-to-end for the first time since Task 1 (#1752) started. Only Task 4 (#1755, dropping the now-unused `rounds`/`time`/`interval` columns and `exercises.workout_id` — a pure schema cleanup with no behavior change) remains in the epic.


app/services/cf_wod/workout_parser.rb
```diff
@@ -110,31 +111,31 @@ def build_exercise_lines(workout, parts)
...
       end
 
       exercise_lines
     end
 
-    def build_exercise_line(workout, line, position:, segment: nil)
+    # A schemed part (rounds/time-window) becomes its own named/scoped segment; a plain,
+    # unschemed run of top-level exercise lines becomes an implicit segment wrapping just that run.
+    def build_part_segment(workout, part, position)
+      return workout.segments.build(position: position) unless part[:segment]
```
Yanchek99: Parsed workouts still keep the header-level scheme on `Workout` (`time`, `rounds`, or `interval` from `WorkoutFormatClassifier`), but after this PR `amrap?`, `interval?`, `set_based_lifting?`, `reps_from_interval`, and related logic all read from `governing_segment`. For a parsed `"21-15-9 reps for time"` workout this builds an unschemed sole segment, so `workout.interval?` becomes false and interval rep calculation/display changes; AMRAP/EMOM/rounds-for-time imports have the same shape. The parser needs to move the applicable header scheme onto the segment it creates, or the classifier/parser contract needs to return segment attributes directly.

app/models/concerns/workout_scoring.rb
```diff
@@ -25,7 +25,7 @@ def calculated_lifting_score?
   def exercises_for_log_recording
     return exercises unless set_based_lifting?
 
-    rounds.times.flat_map { top_level_exercises }
+    rounds.times.flat_map { governing_segment_exercises }
```
Yanchek99: This repeats `workout.rounds`, but `set_based_lifting?` is now based on `governing_segment.rounds?`. Once rounds live only on the segment, `workout.rounds` can be nil here, causing set-based lifting logs to build no repeated recording rows even though the workout is recognized as set-based. Use the governing segment's rounds count here and cover it with a segment-owned-rounds test.

app/models/workout.rb
```diff
@@ -32,12 +31,24 @@ def self.search_by_name(name)
...
+
+    schemed = segments.select(&:schemed?)
+    schemed.sole if schemed.one?
+  end
+
   def rounds_for_time?
-    rounds.present? && time.blank? && interval.blank?
+    return governing_segment.rounds? || !governing_segment.schemed? if governing_segment
+
+    segments.many?
```
Yanchek99: Returning true for every multi-segment workout without a governing segment makes segmented total-reps workouts also look like rounds-for-time. `workout_objective` checks `rounds_for_time?` before `segmented_total_reps?`, so the new timed-block example added in the tests would render as a for-time objective instead of total reps. This should exclude the timed-block/rep-scored shape (or otherwise make the predicates mutually exclusive) and add a helper/model assertion for the display path.

Yanchek99 (commented):
Review notes: I found a few segment-unification regressions that look independent of the documented Task 5/Task 6 follow-ups. The main issue is that new predicates/scoring now read schemes from `governing_segment`, but the parser and set-based logging paths still leave or read scheme data on `Workout`, so imported/logged workouts can be misclassified or under-recorded.

app/services/cf_wod/workout_parser.rb
Yanchek99: Confirmed and fixed in facff67. `WorkoutParser` now computes `header_scheme(attrs)` (rounds/time*60/interval from the classifier) and threads it onto the sole unscoped segment (`build_part_segment`), with `validate_scheme_applies_once!` raising `UnparseableError` if there's more than one unscoped run — mirroring `db/migrate/20260713120000_backfill_segments_for_top_level_exercises.rb`'s own `reject_ambiguous_runs!` precedent for the same shape. Verified against the existing corpus tests (Fran's interval, EMOM's rounds+time_seconds, set-based-lifting's rounds, and the CFJ-181202/time-windowed cases, which correctly stay untouched since PartSplitter already owns their scheme).

app/models/concerns/workout_scoring.rb
Yanchek99: Confirmed and fixed in facff67. `exercises_for_log_recording` now reads `governing_segment.rounds` instead of the workout's own (vestigial) `rounds` column, consistent with `set_based_lifting_structure?`'s existing `governing_segment&.rounds?`.

app/models/workout.rb
Yanchek99: Confirmed and fixed in facff67. `rounds_for_time?`'s no-governing-segment fallback is now `segments.many? && !segmented_total_reps?`, so the rep-scored timed-block shape falls through to `segmented_total_reps?` in `workout_objective` instead of being claimed by `rounds_for_time?` first. Verified against the existing 'a rep-scored AMRAP split across timed blocks' test and the CF-260622 corpus test (4 real time-windowed segments, unaffected since that shape was never hitting the buggy branch to begin with).

Yanchek99 (commented):
Second pass on `facff67`: no new actionable issues found.

Verified the prior threads are addressed:
- Parser now carries header schemes onto built segments.
- Set-based logging repeats `governing_segment.rounds`.
- `rounds_for_time?` no longer masks `segmented_total_reps?`.

The remaining risks I see are the PR body’s documented sibling-task dependencies (#1753 seeds and #1756 show/edit/controller paths), not new review findings in this PR.

app/helpers/workouts_helper.rb
```diff
@@ -11,10 +11,10 @@ def workout_objective(workout)
   def clock_objective(workout)
     return total_reps_clock_objective(workout) if workout.segmented_total_reps?
     return amrap_objective(workout) if workout.amrap?
-    return "EMOM #{workout.time}" if workout.emom?
+    return "EMOM #{workout.governing_segment.time_seconds / 60}" if workout.emom?
     return timed_rounds_objective(workout) if workout.timed_rounds?
 
-    "#{workout.interval} for #{workout.score_measurement}"
+    "#{workout.governing_segment.interval_scheme} for #{workout.score_measurement}"
```
Yanchek99, line 18: This now crashes for a valid segment-less workout because `governing_segment` can still be nil. `Workout` has no validation requiring segments, and this PR's controller test still creates a workout with only `name`, `notes`, and `score_type`; calling `workout_objective` on that record falls through to `clock_objective` and raises `NoMethodError` on `.interval_scheme`. Either make a segment required/defaulted on create, or keep this helper nil-safe for the existing permitted state.

app/views/workouts/_segment_fields.html.slim
```diff
@@ -23,20 +23,20 @@
         - unless implicit_wrapper
```
Yanchek99: After removing the workout-level scheme inputs from `_form`, this guard leaves persisted implicit-wrapper segments with no way to edit `name`, `rounds`, `time_seconds`, `interval_scheme`, `rest_seconds`, or `notes`. Before this PR, the hidden segment scheme was still editable through the workout-level fields; now a user would have to delete/recreate the segment to turn a flat migrated/backfilled workout into an AMRAP/interval/rounds segment. Segment-only editing should expose these controls or provide an explicit conversion path.

Yanchek99 (commented):
Thanks for pushing this through. The broad segment-only direction looks coherent and the high-risk parser/form/test areas have real coverage, but I found a couple of regressions that should be fixed before merge.

One broader issue that does not attach cleanly to a changed line: `exercises.segment_id` is still nullable in `db/schema.rb` and the legacy partial index for `segment_id IS NULL` is still present, even though `Exercise` now declares a required `belongs_to :segment` and the PR summary says every exercise requires a segment. If the schema cleanup is intentionally deferred to Task 4, that task should explicitly include `segment_id` null enforcement and removing the obsolete null-segment index; otherwise this PR leaves the main invariant enforceable only by Rails validations.

app/helpers/workouts_helper.rb
Yanchek99: Confirmed and already fixed in c020362 (before this comment was visible to me) — same bug my own whole-branch-review pass caught independently. `clock_objective` now returns `"For #{workout.score_measurement}"` when `governing_segment.blank?` (app/helpers/workouts_helper.rb:16), before the unguarded `.interval_scheme` call. Covered by a new helper test (`workouts_helper_test.rb`: `'renders a segment-less workout with a generic fallback objective'`).

app/views/workouts/_segment_fields.html.slim
Yanchek99: Confirmed and fixed in 52002a2. Verified live against Murph's segment that `implicit_workout_part?` evaluates true precisely because `workout.rounds/time/interval` are now null (Task 6's fixture rewrite + the already-merged seed rewrite both stop setting them) — so the guard was hiding scheme controls for every plain unschemed/unnamed persisted segment, not just genuine migration-backfilled wrappers, with no remaining path to set a scheme since the workout-level inputs are gone. Removed the `unless implicit_wrapper` guard entirely — the segment's own fields are now the only way to edit any segment's scheme, so there's no case where they should stay hidden. Added a system test (verified RED without the fix via git stash) confirming Murph's Rounds field is present and editable after expanding.

app/views/workouts/_segment_fields.html.slim
```diff
@@ -20,23 +19,22 @@
...
-            .col-12.col-md-6.col-lg-4 = f.input :interval_scheme, placeholder: '21-15-9...', input_html: { data: { 'segment-card-target': 'intervalInput' } }
-          .row.g-2
-            .col-12.col-md-4 = f.input :rest_seconds, input_html: { data: { 'segment-card-target': 'restInput' } }
-            .col-12.col-md-8 = f.input :notes, input_html: { data: { 'segment-card-target': 'notesInput' } }
+        .row.g-2
+          .col-12.col-lg-4 = f.input :name, input_html: { data: { 'segment-card-target': 'nameInput' } }
+          .col-6.col-md-3.col-lg-2 = f.input :rounds, input_html: { data: { 'segment-card-target': 'roundsInput', 'ladder-fields-target': 'rounds', action: 'input->ladder-fields#toggle' } }
+          .col-6.col-md-3.col-lg-2 = f.input :time_seconds, as: :string, placeholder: '20:00 or 1200', input_html: { data: { 'segment-card-target': 'timeInput', 'ladder-fields-target': 'time', action: 'input->ladder-fields#toggle' } }
+          .col-12.col-md-6.col-lg-4 = f.input :interval_scheme, placeholder: '21-15-9...', input_html: { data: { 'segment-card-target': 'intervalInput', 'ladder-fields-target': 'interval', action: 'input->ladder-fields#toggle' } }
```
Yanchek99, line 26: Now that `_segment_fields` always renders and submits `interval_scheme`, saving a timed-round segment with the interval field blank persists `""`, not `nil`. This predicate then returns false, so a segment with `rounds` and `time_seconds` falls through to the interval fallback and renders an objective like `" for rep"`/`" for time"`. I verified this in the PR worktree with `s.update!(interval_scheme: '')`: `s.timed_rounds?` was false and `workout_objective` returned `" for rep"`. Please normalize blank `interval_scheme` to nil or change this check to `interval_scheme.blank?`, and add a regression that submits the workout form rather than only clicking the client-side `Done` button.

Yanchek99 (commented):
Re-reviewed the updated head (`52002a2`). The previous helper nil-crash and flat-segment editability issues are addressed, and the focused regression suite passed locally after adding built assets to the temporary worktree (`31 runs, 71 assertions`). I found one remaining persisted-form regression below.

app/views/workouts/_segment_fields.html.slim
Yanchek99: Confirmed and fixed in 66e4222 — reproduced your exact repro (`s.update!(interval_scheme: '')`) before fixing. `Segment#timed_rounds?` checked `interval_scheme.nil?`, inconsistent with `#rounds?`/`#amrap?` right above it, which both already use `.blank?`. Changed it to match. Added both a model test for the predicate directly and, per your specific ask, a controller test that submits a real `patch workout_url` with `segments_attributes` carrying `interval_scheme: ''` (not just the client-side Done-button path) — both verified RED without the fix via git stash.

Yanchek99 (commented):
Final re-review of updated head `66e4222`: no remaining actionable findings. The submitted-blank `interval_scheme` issue is addressed by treating blank values as absent in `Segment#timed_rounds?`, with both model-level coverage and a persisted controller update regression. CI is green, so I did not do a broader local rerun beyond verifying the specific repro path.

PR #1767, opened 2026-07-16
Fix unfollow for legacy program subscriptions

Yanchek99:
## Summary

- Allow legacy roleless program subscriptions to render the Unfollow action
- Backfill roleless subscriptions to the athlete role and add an athlete default for future rows
- Correct the historical subscription role enum migration typo from `athelete` to `athlete`
- Add regression coverage for the program action UI and role validation

## Root Cause

The old subscription role enum migration mapped the misspelled string `athelete`, so existing `athlete` subscriptions could become `role = nil`. Those rows still made `Current.user.subscribed?` return true, but `can? :unsubscribe` only allowed explicit `athlete` or `coach` roles, so the Unfollow button did not render.

## Validation

- `rvm 4.0.5@wod-tracker do bundle exec rails test test/models/subscription_test.rb test/controllers/turbo_conventions_test.rb`
- `rvm 4.0.5@wod-tracker do bundle exec rubocop app/models/ability.rb app/models/subscription.rb db/migrate/20250311180449_change_role_to_enum_in_subscription.rb db/migrate/20260716180000_backfill_roleless_subscriptions.rb test/controllers/turbo_conventions_test.rb test/models/subscription_test.rb`
- `git diff --check`

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -59,6 +59,7 @@ class Ability
   end
 
   def unsubscribable_role?(subscription)
+    # role.nil? supports legacy rows until the backfill has run everywhere.
     subscription.athlete? || subscription.coach? || subscription.role.nil?
   end
 end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -33,7 +33,7 @@ class Ability
     can [:create, :subscribe], Program
     can [:update, :destroy], Program, subscriptions: { user_id: user.id, role: :owner }
     cannot :subscribe, Program, subscriptions: { user_id: user.id }
-    can :unsubscribe, Program, subscriptions: { user_id: user.id, role: [:athlete, :coach] }
+    can(:unsubscribe, Program) { |program| unsubscribable_program?(program, user) }
 
     can :create, Schedule do |schedule|
       user.programs.manageable.include?(schedule.program)
@@ -48,5 +48,17 @@ class Ability
     cannot :subscribe, Program do |program|
       program.subscriptions.any? { |s| s.user_id == user.id }
     end
+  end
+
+  private
+
+  def unsubscribable_program?(program, user)
+    program.subscriptions.any? do |subscription|
+      subscription.user_id == user.id && unsubscribable_role?(subscription)
+    end
+  end
+
+  def unsubscribable_role?(subscription)
+    subscription.athlete? || subscription.coach? || subscription.role.nil?
   end
 end
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -4,5 +4,6 @@ class Subscription < ApplicationRecord
   belongs_to :program
   belongs_to :user
 
+  validates :role, presence: true
   validates :program, uniqueness: { scope: :user }
 end
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -4,6 +4,7 @@ class Subscription < ApplicationRecord
   belongs_to :program
   belongs_to :user
 
+  # Existing nil roles are repaired by BackfillRolelessSubscriptions before this validation matters.
   validates :role, presence: true
   validates :program, uniqueness: { scope: :user }
 end
--- a/db/migrate/20250311180449_change_role_to_enum_in_subscription.rb
+++ b/db/migrate/20250311180449_change_role_to_enum_in_subscription.rb
@@ -4,13 +4,13 @@ class ChangeRoleToEnumInSubscription < ActiveRecord::Migration[8.0]
     add_column :subscriptions, :role_int, :integer
 
     # Step 2: Map existing string values to integers
-    role_mapping = { 'owner' => 0, 'coach' => 1, 'athlete' => 2 }
+    role_mapping = { 'owner' => 0, 'coach' => 1, 'athelete' => 2 }
 
     Subscription.reset_column_information
     Subscription.find_each do |subscription|
-      next unless role_mapping.key?(subscription.role)
-
-      subscription.update_columns(role_int: role_mapping[subscription.role]) # rubocop:disable Rails/SkipsModelValidations
+      if role_mapping.key?(subscription.role)
+        subscription.update_columns(role_int: role_mapping[subscription.role])
+      end
     end
 
     # Step 3: Remove old string column and rename new column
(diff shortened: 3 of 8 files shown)


Yanchek99:
## Code Review

### Strengths
- Root cause correctly diagnosed and fixed at the right layer: `unsubscribable_role?` explicitly enumerates the three legitimate states (athlete, coach, legacy nil) rather than papering over it with a broader `cannot`/`nil?` hack. The `owner` role is correctly excluded, matching prior behavior.
- Ability refactor to block form (`can(:unsubscribe, Program) { |program| ... }`) is consistent with the existing style in this same file (`:create, Schedule` and the admin `cannot` blocks already use blocks); confirmed `:unsubscribe` is only ever checked via instance-level `can?`/`load_and_authorize_resource`, never `accessible_by`, so the switch away from hash conditions doesn't break any SQL-scoping use.
- `validates :role, presence: true` closes the hole going forward, backed by a real, focused unit test.
- Backfill migration uses an isolated `MigrationSubscription` class scoped to the table rather than the live `Subscription` model — good insulation from future model changes.
- New regression test in `turbo_conventions_test.rb` uses `update_column(:role, nil)` to faithfully simulate the legacy corrupted state and asserts against the actual rendered Unfollow link — a real behavioral test, not a mock.

### Issues

#### Important (Should Fix)
- `db/migrate/20260716180000_backfill_roleless_subscriptions.rb:6` — `MigrationSubscription.where(role: nil).update_all(role: 2)` runs as one unbatched UPDATE. On a large production table this can hold row/table locks for a long time, and `role` has no index, so it's at minimum a full table scan. Recommend `in_batches.update_all` regardless of current table size.
- Editing already-applied migration `db/migrate/20250311180449_change_role_to_enum_in_subscription.rb` (typo fix) has no effect on any environment where it already ran — `schema_migrations` won't re-execute it. It only helps hypothetical fresh `db:migrate` replays that don't use `db:schema:load`. Low risk since it's only a data-mapping string, but it's editing history rather than adding a new migration — worth confirming this was an intentional, accepted tradeoff.
- Adding `default: 2` to `subscriptions.role` retroactively changes what any fixture/seed row gets when `role` is omitted (previously NULL, now `athlete`). Doesn't break the two existing test files (both explicitly set role before asserting), but worth flagging for future fixtures/seeds that omit `role`.

#### Minor (Nice to Have)
- Migration `down` reverts only the column default, not the data backfill — previously-nil rows that got set to `athlete` stay `athlete` after rollback. Standard for backfills, but undocumented in the migration.
- Deploy-ordering assumption: `validates :role, presence: true` will reject any `.save` (not just `.destroy`) on a legacy nil-role row until the backfill migration has run. Fine under normal migrate-then-boot deploys, but worth a one-line comment given it's a hard validation added alongside a data-fix migration.

### Recommendations
Consider adding a short comment in `unsubscribable_role?` noting that `role.nil?` specifically exists to support pre-backfill legacy rows, so a future reader isn't tempted to remove it once all rows are guaranteed non-null.

### Assessment

**Ready to merge?** With fixes (batch the backfill `update_all`; confirm the historical-migration edit was a deliberate choice).

**Reasoning:** The core fix is correct, well-tested, and consistent with existing codebase conventions; the only material production-readiness gap is the unbatched backfill update, which is a should-fix rather than a blocker at this app's likely scale.

---

Yanchek99:
Addressed the actionable code feedback in `f179a75`:

- Batched the roleless subscription backfill with `in_batches.update_all`.
- Added a comment on the legacy `role.nil?` unsubscribe path.
- Documented that the migration rollback only restores the schema default and does not reverse the data backfill.

On the edited historical migration: that was intentional. It will not affect environments that already ran the migration, but it keeps a replayed migration path from preserving the original `athelete` typo. The new backfill migration is still what repairs already-migrated environments.

Validation rerun:

- `rvm 4.0.5@wod-tracker do bundle exec rails test test/models/subscription_test.rb test/controllers/turbo_conventions_test.rb`
- `rvm 4.0.5@wod-tracker do bundle exec rubocop app/models/ability.rb app/models/subscription.rb db/migrate/20250311180449_change_role_to_enum_in_subscription.rb db/migrate/20260716180000_backfill_roleless_subscriptions.rb test/controllers/turbo_conventions_test.rb test/models/subscription_test.rb`
- `git diff --check`

Yanchek99:
## Re-review after `f179a75`

Verified `f179a75` addresses the prior review:
- Backfill migration now batches (`in_batches.update_all`) — Important issue resolved.
- `role.nil?` explanatory comment added to `unsubscribable_role?` — Recommendation addressed.
- `down` now documents that the data backfill isn't reversed — Minor issue addressed.

Ran the full relevant suite (`subscription_test.rb`, `turbo_conventions_test.rb`) and RuboCop across all 6 touched files: all green, no offenses.

### Still open (judgment calls, not blockers)
- `db/migrate/20250311180449_change_role_to_enum_in_subscription.rb` — the `athelete` → `athlete` typo fix edits an already-applied migration. Harmless in any environment that already ran it (schema_migrations won't re-run it), but worth a quick confirmation this was a deliberate choice rather than an oversight, since editing history is usually avoided in favor of a new migration.
- `db/schema.rb` / migration — `default: 2` on `subscriptions.role` means any future fixture or seed row that omits `role` now silently gets `athlete` instead of `NULL`. Doesn't break anything today, just worth being aware of.
- `app/models/subscription.rb:7` — `validates :role, presence: true` implicitly depends on the backfill migration having run first (any `.save` on a legacy nil-role row will fail validation until then). No code change needed, just flagging for awareness — a one-line comment here would make the dependency explicit for future readers.

**Assessment: Ready to merge.** No remaining Critical/Important issues.

---

Yanchek99:
Addressed the remaining judgment-call feedback in `6b8f761`:

- Reverted the edit to the already-applied `20250311180449_change_role_to_enum_in_subscription` migration. The final PR diff no longer touches that historical migration; the new backfill migration is the repair path.
- Made `test/fixtures/subscriptions.yml` roles explicit so the current fixtures do not silently depend on the new database default.
- Added a comment at the `Subscription` role validation documenting that existing nil roles are repaired by `BackfillRolelessSubscriptions` before the validation matters.
- Added a migration comment clarifying that omitted roles default to normal `athlete` follows.

Validation rerun after these changes:

- `rvm 4.0.5@wod-tracker do bundle exec rails test test/models/subscription_test.rb test/controllers/turbo_conventions_test.rb`
- `rvm 4.0.5@wod-tracker do bundle exec rubocop app/models/ability.rb app/models/subscription.rb db/migrate/20260716180000_backfill_roleless_subscriptions.rb test/controllers/turbo_conventions_test.rb test/models/subscription_test.rb`
- `git diff --check`

Issue #1751, opened 2026-07-13
Collapse Workout scheme into Segment: tracking issue


Yanchek99:
## Goal

Implement the design in #1750: every `Exercise` always belongs to a `Segment` (a "flat" stretch of exercises is just an unschemed segment); `Workout` drops its own `rounds`/`time`/`interval` columns and keeps only genuinely whole-workout fields (`name`, `score_type`, `team_size`, `ladder_step`, `time_cap_seconds`, `content_key`, `notes`). This removes the duplicated scheme predicates between `Workout`/`Segment` and the shared-position-namespace logic between top-level `Exercise` rows and `Segment` rows.

## Tasks (in dependency order)

- [x] #1752 — Backfill implicit segments for top-level exercises (data-only migration)
- [x] #1753 — Rewrite seed files to always wrap exercises in segments
- [x] #1754 — Delegate workout scheme predicates to a governing segment (core model/concern rewiring)
- [x] #1756 — Segment-only workout show page, builder form, and controller params
- [ ] #1755 — Drop workout scheme columns and exercises.workout_id (final schema cleanup)
- [x] #1757 — Fixtures, controller tests, and full regression pass

## Dependency graph

- #1752 (Task 1) and #1753 (Task 2) have no dependency on each other or anything else — safe to start immediately, in parallel.
- #1754 (Task 3) depends on #1752 (Task 1) merged and deployed — the data must already be backfilled before the model stops allowing `segment_id: nil`.
- #1756 (Task 5) depends on #1754 (Task 3) — needs `Workout#governing_segment`/`Segment#schemed?` and `workout.segments` as the one sequence.
- #1755 (Task 4) depends on **both** #1754 (Task 3) **and** #1756 (Task 5) — the workout builder form still binds directly to `Workout#rounds`/`time`/`interval` until Task 5 lands, so the columns can't be dropped until both the model layer (3) and the form (5) have stopped using them. (This is stricter than a simple "depends on Task 3" — worth calling out since it's easy to schedule Task 4 too early.)
- #1757 (Task 6) depends on #1754 (Task 3) and #1756 (Task 5) — final fixture/test rewrite and full-suite regression pass; best run last since it's the epic's overall gate, though it could technically start once 3 and 5 are in.

## Notes

- This is a real schema migration touching 4 seed files (~350 workout definitions), 6+ model/concern files, and 5+ views — implemented as 6 separate tasks, not one PR.
[...]

PR #1768, opened 2026-07-16
Segment unification Task 4: Drop workout scheme columns and exercises.workout_id

### LINKED ISSUE #1755: Segment unification Task 4: Drop workout scheme columns and exercises.workout_id

Yanchek99:
## Summary
- Adds a DB-level `NOT NULL` on `exercises.segment_id` (safe — verified 0 exercises with a nil `segment_id`) and drops `workouts.rounds`/`time`/`interval` and `exercises.workout_id` entirely, along with the model code that only existed to keep `workout_id` in sync.
- While getting the full suite green, also fixed 4 additional live/test call sites the issue's brief didn't enumerate: `Exercise#workout` internal callers (via a private `delegate :workout, to: :segment`), `CfWod::WorkoutParser` passing scheme keys straight to `Workout.new`, a dead `ImplicitWorkoutPart` comparison against the removed workout scheme, and remaining test/fixture references to the dropped columns (including deleting an 8-test file that became permanently unexecutable once the columns it exercised were removed).

Closes #1755 (part of #1751).

## Test plan
- [x] `bin/rails test` — 371 runs, 1014 assertions, 0 failures, 0 errors, 0 skips (baseline was 379/1047; delta is exactly the removed obsolete migration test file)
- [x] Migration round-tripped (`db:migrate` / `db:rollback` / `db:migrate`) with byte-identical `db/schema.rb` before/after a rubocop-driven restructuring
- [x] `bundle exec rubocop` clean on all touched files
- [x] Task-scoped review: spec compliant, Approved
- [x] Final whole-branch review: Ready to merge, no Critical/Important issues


Yanchek99 (commented):
I found one migration-safety issue worth addressing before merge:

**Important:** `db/migrate/20260713150000_drop_workout_scheme_and_exercise_workout_id.rb:3` changes `exercises.segment_id` to `NOT NULL` before dealing with any remaining null rows. The previous segment backfill explicitly tolerated malformed/legacy exercises that had no `workout_id`, which means a production database containing even one of those rows will fail this deploy migration at `change_column_null` before the column cleanup can run. I would add a preflight/backfill/delete strategy for `Exercise.unscoped.where(segment_id: nil)` before the constraint, and cover that path with a migration test so this migration either repairs known-safe rows or fails with a deliberate, explanatory error.

I also ran the focused PR test set against `origin/pr/1768` in a temporary worktree:

`rvm 4.0.5@wod-tracker do bundle exec rails test test/migrations/drop_workout_scheme_and_exercise_workout_id_test.rb test/models/workout_ascending_ladder_test.rb test/models/log_test.rb test/models/workout_fingerprint_test.rb test/models/workout_test.rb test/services/cf_wod/workout_parser_test.rb test/services/cf_wod/workout_parser_corpus_test.rb`

Result: 76 runs, 288 assertions, 0 failures, 0 errors.

Yanchek99:
Fixed in ab01836. Verified the claim first: the migration test file deleted in this PR (`backfill_segments_for_top_level_exercises_test.rb`) had a test named `'ignores malformed top-level exercises without a workout id'` that inserted an `Exercise` with no `workout_id` via `Exercise.insert!` (bypassing validations) and asserted `BackfillSegmentsForTopLevelExercises#up` does *not* raise on it. That migration's `up` computes `MigrationExercise.where(segment_id: nil).distinct.pluck(:workout_id).compact` — the `.compact` silently drops nil `workout_id`s from the backfill list, so any exercise with neither `workout_id` nor `segment_id` set is permanently skipped. Confirmed this was a deliberate, previously-tested tolerance, not a hypothetical.

Added `guard_against_orphaned_exercises!` to `db/migrate/20260713150000_drop_workout_scheme_and_exercise_workout_id.rb`, run at the top of `up`. It checks `Exercise.unscoped.where(segment_id: nil).count` and raises a message pointing at the query to investigate, rather than attempting an automated repair — there's no way to know from inside the migration what a stray orphaned row should be attached to, so failing loud and explicit is safer than guessing. Added a migration test that reconstructs the orphaned-row scenario (temporarily relaxing the NOT NULL constraint, inserting via `Exercise.insert!`, asserting the guard raises with an actionable message) since the schema's NOT NULL constraint makes it otherwise unconstructable post-migration.

Full suite: 372 runs, 1017 assertions, 0 failures, 0 errors, 0 skips. Rubocop clean on both touched files.

PR #1771, opened 2026-07-17
Use WorkoutExtraction::LlmParser in ScrapeCfWodJob instead of the heuristic parser

### LINKED ISSUE #1733: LLM WOD extraction Task 3: Wire into ScrapeCfWodJob
## Goal

Swap `CfWod::WorkoutParser` for `WodExtraction::LlmParser` in the scraper job's call site.

Depends on: Task 2 (`WodExtraction::LlmParser`) merged to master, AND the in-progress `scrape-crossfit-wod-job-1677` branch (PR for #1677) merged to master. **This task cannot land before both are true** — if `scrape-crossfit-wod-job-1677` merges first, rebase this task's branch on top of master after that merge; if Task 2 merges first, this task waits for `scrape-crossfit-wod-job-1677` to merge, then applies the change directly on master.

**Files:**
- Modify: `app/jobs/scrape_cf_wod_job.rb`
- Modify: `test/jobs/scrape_cf_wod_job_test.rb`

**Interfaces:**
- Consumes: `WodExtraction::LlmParser.call(text)` and `WodExtraction::LlmParser::ExtractionError` (Task 2).

## Steps

- [ ] **Step 1: Confirm the current state of the call site**

```bash
git log --oneline master..scrape-crossfit-wod-job-1677 -- app/jobs/scrape_cf_wod_job.rb 2>/dev/null || true
cat app/jobs/scrape_cf_wod_job.rb
```

Confirm the file matches this shape (as of the branch this was scoped against):

```ruby
class ScrapeCfWodJob < ApplicationJob
  queue_as :default

  retry_on CfWod::Fetcher::FetchError, wait: :polynomially_longer, attempts: 3 do |job, error|
    WodImport.log_failure!(job.arguments.first, error.message)
  end

  retry_on CfWod::Fetcher::UnrecognizedTemplateError, attempts: 1 do |job, error|
    WodImport.log_failure!(job.arguments.first, error.message)
  end

  def self.default_date
    ActiveSupport::TimeZone['America/Los_Angeles'].today + 1
  end

  def perform(date = self.class.default_date)
    self.arguments = [date]

    page = CfWod::Fetcher.call(date)
    return if page.rest_day?

    workout = CfWod::WorkoutParser.call(page)
    workout = persist(workout)
    Program.find_by!(name: 'Crossfit.com')
           .schedules.find_or_initialize_by(posted_at: date)
           .update!(workout: workout)
    WodImport.clear!(date)
  rescue CfWod::WorkoutParser::UnparseableError, ActiveRecord::ActiveRecordError => e
    WodImport.log_failure!(date, e.message, raw_text: page&.body_text)
  end

  private

  def persist(workout)
    return workout if workout.persisted?

    workout.save!
    workout.absorb_duplicate!
  end
end
```

If the file has diverged from this (e.g. the branch changed error handling), adapt the edit below to the real current content rather than blind-applying the diff.

- [ ] **Step 2: Swap the parser call and rescued exception class**

```ruby
    workout = WodExtraction::LlmParser.call(page.body_text)
    workout = persist(workout)
    Program.find_by!(name: 'Crossfit.com')
           .schedules.find_or_initialize_by(posted_at: date)
           .update!(workout: workout)
    WodImport.clear!(date)
  rescue WodExtraction::LlmParser::ExtractionError, ActiveRecord::ActiveRecordError => e
    WodImport.log_failure!(date, e.message, raw_text: page&.body_text)
```

Everything else in the file (the `retry_on` blocks, `default_date`, `persist`) is unchanged.

- [ ] **Step 3: Update the job's test to stub the LLM parser instead of exercising the heuristic one**

```bash
cat test/jobs/scrape_cf_wod_job_test.rb
```

Find every test that currently drives real WOD-page fixtures through `CfWod::WorkoutParser` to produce a persisted `Workout`. Replace the parser dependency with a stub, e.g.:

```ruby
WodExtraction::LlmParser.stub(:call, workouts(:some_fixture)) do
  ScrapeCfWodJob.perform_now(date)
end
```

For the failure-path test (currently likely feeding an unparseable fixture through the real heuristic parser to trigger `WodImport.log_failure!`), stub `WodExtraction::LlmParser.call` to raise `WodExtraction::LlmParser::ExtractionError` instead:

```ruby
WodExtraction::LlmParser.stub(:call, ->(_text) { raise WodExtraction::LlmParser::ExtractionError, "boom" }) do
  ScrapeCfWodJob.perform_now(date)
end
assert WodImport.exists?(wod_date: date)
```

Match the exact test names, fixtures, and assertion style already in the file — this step's exact diff depends on that file's current content, which is why it isn't fully specified here.

- [ ] **Step 4: Run the job's tests**

```bash
bin/rails test test/jobs/scrape_cf_wod_job_test.rb
```

Expected: all pass, none of them making a real Anthropic API call (verify no `WebMock::NetConnectNotAllowedError` or similar).

- [ ] **Step 5: Commit**

```bash
git add app/jobs/scrape_cf_wod_job.rb test/jobs/scrape_cf_wod_job_test.rb
git commit -m "Use WodExtraction::LlmParser in ScrapeCfWodJob instead of the heuristic parser"
```

## Verification

- `bin/rails test test/jobs/scrape_cf_wod_job_test.rb` passes
- `bundle exec rubocop app/jobs/scrape_cf_wod_job.rb` passes
- Grep confirms no remaining reference to `CfWod::WorkoutParser` in `app/jobs/scrape_cf_wod_job.rb`: `grep -n CfWod::WorkoutParser app/jobs/scrape_cf_wod_job.rb` returns nothing

---
Part of #1730

Yanchek99:
## Summary
- Swaps `CfWod::WorkoutParser` for `WorkoutExtraction::LlmParser` at the `ScrapeCfWodJob` call site, per #1733.
- Note: #1733's original text described a stale interface (`WodExtraction::LlmParser.call(text)`, single `ExtractionError`). This PR is built against the actual current interface on master: `WorkoutExtraction::LlmParser.call(text, date:, logger: nil)`, which raises both `ExtractionError` and `UnrepresentableWorkoutError` — both now rescued alongside `ActiveRecord::ActiveRecordError`.
- Updates `test/jobs/scrape_cf_wod_job_test.rb` so every test that reaches the parser stubs `WorkoutExtraction::LlmParser.call` instead of driving real HTML fixtures through the heuristic parser (no test hits the real Anthropic API).
- Adds a hand-rolled `stub_llm_parser` test helper since this project's minitest (6.0.6) no longer ships `Object#stub` (split into an unused `minitest-mock` gem), and adds the test file to `.rubocop.yml`'s `Metrics/ClassLength` exclude list (matching 8 other test files already excluded there) since the helper pushed the class past the 100-line default.

## Test plan
- [x] `bin/rails test test/jobs/scrape_cf_wod_job_test.rb` — 9/9 passing, pristine output, no real Anthropic API calls
- [x] `bundle exec rubocop app/jobs/scrape_cf_wod_job.rb test/jobs/scrape_cf_wod_job_test.rb` — 0 offenses
- [x] `grep -n CfWod::WorkoutParser app/jobs/scrape_cf_wod_job.rb test/jobs/scrape_cf_wod_job_test.rb` — no matches
- [x] Task-scoped subagent review: approved (spec compliant, no Critical/Important issues)
- [x] Final whole-branch subagent review: approved, ready to merge (no Critical/Important issues)

Closes #1733

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -26,5 +26,19 @@ module ActiveSupport
       stub_request(:get, %r{\Ahttps://www\.crossfit\.com/workout/#{legacy_path}})
         .to_return(status: 301, headers: { 'Location' => "/#{slug}" })
     end
+
+    # Object#stub (from minitest/mock) isn't available here -- minitest 6 dropped mock.rb into a
+    # separate minitest-mock gem this app doesn't depend on -- so stub WorkoutExtraction::LlmParser.call
+    # by hand: swap in a replacement singleton method for the duration of the block, then restore the
+    # original. `result` may be a plain value to return (e.g. a persisted Workout fixture) or a
+    # callable (e.g. a lambda that raises) to invoke with the same arguments the job passes.
+    def stub_llm_parser(result)
+      original_call = WorkoutExtraction::LlmParser.method(:call)
+      responder = result.respond_to?(:call) ? result : ->(*) { result }
+      WorkoutExtraction::LlmParser.define_singleton_method(:call) { |*args, **kwargs| responder.call(*args, **kwargs) }
+      yield
+    ensure
+      WorkoutExtraction::LlmParser.define_singleton_method(:call, original_call)
+    end
   end
 end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -55,6 +55,7 @@ Metrics/ClassLength:
     - test/system/workout_drag_ordering_test.rb
     - test/system/workout_sortable_card_inputs_test.rb
     - test/controllers/workouts_controller_test.rb
+    - test/jobs/scrape_cf_wod_job_test.rb
 
 Metrics/MethodLength:
   CountComments: false
--- a/app/jobs/scrape_cf_wod_job.rb
+++ b/app/jobs/scrape_cf_wod_job.rb
@@ -2,13 +2,23 @@ class ScrapeCfWodJob < ApplicationJob
   queue_as :default
 
   # Keyed lookup of the two available extraction strategies, so `perform` can pick one at
-  # runtime (default :llm) without hardcoding either parser's call signature inline. :heuristic
-  # is kept available as a fallback -- e.g. `ScrapeCfWodJob.perform_later(date, :heuristic)` --
-  # in case the LLM parser needs to be bypassed for a given date.
+  # runtime without hardcoding either parser's call signature inline. The primary strategy
+  # defaults to Rails.application.config.workout_parser, overridable per call (e.g.
+  # `ScrapeCfWodJob.perform_later(date, :heuristic)`).
   PARSERS = {
     llm: ->(page, date) { WorkoutExtraction::LlmParser.call(page.body_text, date: date) },
     heuristic: ->(page, _date) { CfWod::WorkoutParser.call(page) }
   }.freeze
+
+  # Each parser's own "couldn't extract a workout" exception class(es) -- extract_workout
+  # rescues exactly these for the primary strategy before retrying with its fallback.
+  PARSER_ERRORS = {
+    llm: [WorkoutExtraction::LlmParser::ExtractionError, WorkoutExtraction::LlmParser::UnrepresentableWorkoutError].freeze,
+    heuristic: [CfWod::WorkoutParser::UnparseableError].freeze
+  }.freeze
+
+  # The other strategy to fall back to when the primary one fails to extract a workout.
+  FALLBACKS = { llm: :heuristic, heuristic: :llm }.freeze
 
   retry_on CfWod::Fetcher::FetchError, wait: :polynomially_longer, attempts: 3 do |job, error|
     WorkoutImport.log_failure!(job.arguments.first, error.message)
@@ -35,13 +45,13 @@ class ScrapeCfWodJob < ApplicationJob
   # retry that crosses local midnight in America/Los_Angeles would silently target a different
   # day's workout than the attempt before it -- and would silently fall back to the default parser
   # on retry if `parser` were left unpinned.
-  def perform(date = self.class.default_date, parser = :llm)
+  def perform(date = self.class.default_date, parser = Rails.application.config.workout_parser)
     self.arguments = [date, parser]
 
     page = CfWod::Fetcher.call(date)
     return if page.rest_day?
 
-    workout = PARSERS.fetch(parser).call(page, date)
+    workout = extract_workout(page, date, parser)
     workout = persist(workout)
     Program.find_by!(name: 'Crossfit.com')
            .schedules.find_or_initialize_by(posted_at: date)
@@ -55,6 +65,16 @@ class ScrapeCfWodJob < ApplicationJob
   end
 
   private
+
+  # Try the primary parser; if it fails with its own "couldn't extract a workout" error, retry
+  # once with the other parser instead of failing the job outright. A failure from the fallback
+  # itself is not rescued here -- it propagates to perform's own rescue clause, which logs the
+  # WorkoutImport failure once both strategies have been tried.
+  def extract_workout(page, date, parser)
+    PARSERS.fetch(parser).call(page, date)
+  rescue *PARSER_ERRORS.fetch(parser)
+    PARSERS.fetch(FALLBACKS.fetch(parser)).call(page, date)
+  end
 
   def persist(workout)
     return workout if workout.persisted?
--- a/app/jobs/scrape_cf_wod_job.rb
+++ b/app/jobs/scrape_cf_wod_job.rb
@@ -30,13 +30,15 @@ class ScrapeCfWodJob < ApplicationJob
     page = CfWod::Fetcher.call(date)
     return if page.rest_day?
 
-    workout = CfWod::WorkoutParser.call(page)
+    workout = WorkoutExtraction::LlmParser.call(page.body_text, date: date)
     workout = persist(workout)
     Program.find_by!(name: 'Crossfit.com')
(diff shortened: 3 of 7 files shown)


Yanchek99:
Added a runtime toggle so the old heuristic parser stays genuinely usable, not just present as dead code: `perform` now takes an optional second positional argument (`parser`, default `:llm`) dispatched through a `PARSERS` lookup. `ScrapeCfWodJob.perform_later(date, :heuristic)` will run the original `CfWod::WorkoutParser` path instead.

- `parser` is positional rather than keyword deliberately: `retry_on` re-invokes via `job.perform(*job.arguments)`, and a plain (non-`ruby2_keywords`-flagged) Hash inside a splatted array doesn't auto-convert to keyword args in Ruby 3+ — a keyword arg here would break retries.
- Both `CfWod::WorkoutParser::UnparseableError` and the two LLM error classes are now rescued, since either parser can run.
- New test covers the `:heuristic` path end-to-end against the real parser (no LLM stub involved), and existing tests/behavior for the `:llm` default are unchanged.

10/10 tests passing, rubocop clean, reviewed by a fresh subagent (approved, no Critical/Important findings).

Yanchek99:
Fixed CI. Root cause: two other test files invoke \`ScrapeCfWodJob\` indirectly and weren't updated when its default parser switched from the heuristic parser to \`WorkoutExtraction::LlmParser\` — they attempted a real Anthropic API call, which CI's WebMock config correctly blocked.

- \`test/jobs/backfill_crossfit_wods_job_test.rb\` — \`BackfillCrossfitWodsJob\` enqueues \`ScrapeCfWodJob\` without a parser arg (defaults to \`:llm\`).
- \`test/tasks/cf_wod_rake_test.rb\` — the \`cf_wod:scrape\` rake task calls \`ScrapeCfWodJob.perform_now\` directly, same issue.

Both now stub the LLM parser the same way \`scrape_cf_wod_job_test.rb\` does. Moved the \`stub_llm_parser\` helper out of \`ScrapeCfWodJobTest\` and into \`test_helper.rb\`'s shared \`ActiveSupport::TestCase\` so all three test files can use it without duplication.

All checks green: Build, rubocop, CodeQL analyze (ruby/js/actions).

Yanchek99:
Replaced the manual per-call toggle with a config-driven default plus automatic fallback, per your suggestion:

- \`config.workout_parser\` (default \`:llm\`) in \`config/application.rb\` now sets the primary extraction strategy — no existing config with that name, verified by grep.
- \`ScrapeCfWodJob\` automatically retries with the *other* parser if the primary one fails to extract a workout (\`:llm\` fails → \`:heuristic\` fallback, and vice versa), before logging a \`WorkoutImport\` failure. Only a double failure (both parsers) results in a logged failure now.
- The per-call \`parser\` argument still works as an override of the primary choice (e.g. for one specific date), it just now defaults from config instead of a hardcoded literal.
- A reviewer subagent caught that the parser-specific exception lists were duplicated in two places (\`PARSER_ERRORS\` and \`perform\`'s rescue clause) — fixed by deriving the rescue clause from \`PARSER_ERRORS.values.flatten\` so there's one source of truth.

Added 2 new tests covering both fallback directions, and updated the existing "unparseable workout" test since it now exercises a genuine double-failure (both parsers fail) rather than a single stubbed one. 25/25 job-level tests passing, 389/389 across the full non-system suite, rubocop clean, CI green.

test/test_helper.rb
```diff
@@ -26,5 +26,19 @@ def stub_cf_wod_redirect(legacy_path, slug)
...
     end
+
+    # Object#stub (from minitest/mock) isn't available here -- minitest 6 dropped mock.rb into a
+    # separate minitest-mock gem this app doesn't depend on -- so stub WorkoutExtraction::LlmParser.call
+    # by hand: swap in a replacement singleton method for the duration of the block, then restore the
+    # original. `result` may be a plain value to return (e.g. a persisted Workout fixture) or a
+    # callable (e.g. a lambda that raises) to invoke with the same arguments the job passes.
+    def stub_llm_parser(result)
+      original_call = WorkoutExtraction::LlmParser.method(:call)
+      responder = result.respond_to?(:call) ? result : ->(*) { result }
```
Yanchek99, line 37: Non-blocking: the plain-value path here ignores the arguments passed to `WorkoutExtraction::LlmParser.call`, so tests using `stub_llm_parser(workouts(:fran))` would still pass if `ScrapeCfWodJob` accidentally passed the page object instead of `page.body_text`, omitted `date:`, or used the wrong date. Since this PR’s key integration point is the LLM parser call signature, consider adding at least one callable stub/assertion that verifies it receives the body text and `date:` keyword.

Yanchek99 (commented):
Reviewed PR 1771 against the stated goal of switching `ScrapeCfWodJob` to `WorkoutExtraction::LlmParser` with parser config/fallback behavior. I did not find any blocking production issues. I left one non-blocking test-coverage comment inline.

Note: I could not run the targeted Rails tests locally in this session because sandboxed RVM is blocked by `/bin/ps`, and the escalation to run PR code unsandboxed was rejected for safety. Review was based on the PR diff, changed files at `41df4da`, and a read-only subagent pass.

test/test_helper.rb
Yanchek99: Confirmed: the plain-value branch (`->(*) { result }`) discards all arguments, so no test verified the call site's actual signature.

Added a test using a callable responder that captures `(text, date:)` and asserts both. Verified it's a real regression guard, not just a shape check: temporarily changed the call site to `date: date + 1`, confirmed this new test fails, then reverted.

Fixed in 94033b3.

PR #1773, opened 2026-07-19
Fix mobile alignment on programs and schedules views

Yanchek99:
## Summary
- Programs index (`app/views/programs/index.html.slim`) only rendered a checkbox icon for subscribed programs, leaving other rows blank and making the list look uneven on mobile. Now every row renders either a checked or empty square icon (`far fa-square` registered in `app/javascript/application.js`).
- Schedules date nav row (`app/views/schedules/index.html.slim`) wrapped the `>` next-page button onto its own line on narrow viewports because the row allowed flex-wrap while the date heading used a fixed `col-auto` width. Added `flex-nowrap` to the row and switched the date heading to `col` so it wraps its own text instead of pushing the button down.

## Test plan
- [x] `bundle exec rails test test/controllers/schedules_controller_test.rb` passes
- [x] Verified in a mobile-width (375px) browser preview: programs list shows a checkbox column on every row; schedules `<`/`>` buttons stay on one row across multiple dates
- [x] Verified desktop-width layout is unchanged for both views

### DIFF
--- a/app/views/programs/index.html.slim
+++ b/app/views/programs/index.html.slim
@@ -12,3 +12,5 @@
           .col.text-end
             - if Current.user.subscribed? program
               i.far.fa-check-square
+            - else
+              i.far.fa-square
--- a/app/views/programs/index.html.slim
+++ b/app/views/programs/index.html.slim
@@ -1,9 +1,10 @@
 .container
-  h1 Programs
-  - if can? :create, Program
-    .my-2.text-end
-      = link_to new_program_path, class: 'btn btn-primary'
-        i.fas.fa-plus
+  .row.align-items-center.my-2
+    h1.col.mb-0 Programs
+    - if can? :create, Program
+      .col-auto
+        = link_to new_program_path, class: 'btn btn-primary'
+          i.fas.fa-plus
   #programs.list-group
     - @programs.each do |program|
       = link_to program, class: "list-group-item list-group-item-action" do
@@ -12,5 +13,3 @@
           .col.text-end
             - if Current.user.subscribed? program
               i.far.fa-check-square
-            - else
-              i.far.fa-square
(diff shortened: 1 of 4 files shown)


app/views/programs/index.html.slim
```diff
@@ -12,3 +12,5 @@
           .col.text-end
             - if Current.user.subscribed? program
               i.far.fa-check-square
+            - else
+              i.far.fa-square
```
Yanchek99: This is not needed please remove.

.gitignore
```diff
@@ -65,4 +65,5 @@ yarn-debug.log*
 !/app/assets/builds/.keep
 
 # AI
-/.claude
\ No newline at end of file
+/.claude
+/.worktrees
```
Yanchek99: Why is this in here.

Yanchek99 (commented):
The programs add button alignment was never fixed

app/views/programs/index.html.slim
Yanchek99: Removed.

.gitignore
Yanchek99: Dropped — rebased this branch onto origin/master so it no longer carries that commit. It's a separate local-master-only change from before this PR started; happy to land it on master separately if you still want it.

Yanchek99:
Addressed the review feedback:

- Removed the `else`/`fa-square` addition on the programs list (and the now-unused `faSquare` import) — reverted to only showing the check icon for subscribed programs.
- Moved the "+" new-program button onto the same row as the "Programs" heading (title left, button right, vertically centered) instead of its own row below — verified on mobile (375px) and desktop.
- Dropped the `.gitignore` `/.worktrees` line from this branch (rebased onto current origin/master) — that was a separate local-only commit from before this PR, not something this branch should carry.

Pushed as `07243ab`.

Issue #1735, opened 2026-07-10
LLM WOD extraction Task 5: Textarea + escape-hatch views


Yanchek99:
## Goal

Build the textarea page (replacing the old blank-form `new.html.slim` as the primary "New Workout" entry point), move the old blank form to a new `new_manual.html.slim`, and link them together as the escape hatch.

Depends on: Task 4 (routes + controller actions) merged.

**Files:**
- Modify: `app/views/workouts/new.html.slim`
- Create: `app/views/workouts/new_manual.html.slim`

**Interfaces:**
- Consumes: `@workout` (set by `WorkoutsController#new`/`#new_manual`/`#extract`, Task 4).
- Consumes: `extract_workouts_path`, `new_manual_workouts_path` (route helpers from Task 4).
- Consumes: `_form.html.slim` (existing, unchanged — the nested-fields form partial).

## Steps

- [ ] **Step 1: Replace `new.html.slim` with the textarea page**

Current content (this is what moves to `new_manual.html.slim` in Step 2):

```slim
.container
  h1 New Workout
  = render 'form', workout: @workout
```

New content:

```slim
.container
  h1 New Workout
  p.text-muted Paste a workout description below and we'll try to structure it for you.
  = simple_form_for :extraction, url: extract_workouts_path, method: :post, html: { class: 'wod-text-form' } do |f|
    .form-group
      = f.input :wod_text, as: :text, label: 'Workout text', input_html: { rows: 8, required: true }
    .form-actions
      button.btn.btn-success type="submit"
        i.fa-solid.fa-wand-magic-sparkles aria-hidden="true"
        span Structure this workout
      = link_to 'Enter manually instead', new_manual_workouts_path, class: 'btn btn-outline-secondary'
```

The form posts a plain `params[:extraction][:wod_text]` — Step 3 below wires the controller's `params.expect(:wod_text)` call from Task 4 to actually read `params[:extraction][:wod_text]`, since `simple_form_for :extraction` (a symbol, not a model) namespaces the input under `extraction`.

- [ ] **Step 2: Create `new_manual.html.slim` with the old content**

```slim
.container
  h1 New Workout
  = render 'form', workout: @workout
```

[...]

PR #1777, opened 2026-07-20
Structured workout view as default, with toggle to paste-text extraction

Yanchek99:
## Summary

Redesigns #1735: instead of the paste-text ("unstructured") entry point being the only way in, the structured nested-fields form stays the default for both `new` and `edit`, with a plain toggle link to the paste-text page and back — matching the routing this repo already settled into (`new` renders the structured form; `new_unstructured`/`extract` were the only escape hatch, previously with no link connecting them).

- `Workout#replace_with_extraction!` — replaces a persisted workout's attributes and all its segments/exercises in memory from a freshly LLM-extracted (unsaved) workout, via the existing `accepts_nested_attributes_for :segments, allow_destroy: true` machinery. Nothing touches the DB until the caller saves.
- `edit_unstructured` (GET) / `re_extract` (PATCH) controller actions + routes, mirroring the existing `new_unstructured`/`extract` pair for an existing workout. `re_extract` calls the model method above and re-renders the (unsaved) edit form, or falls back to the paste-text view with a flash alert on extraction failure.
- `_segment_fields.html.slim` now hides segments marked for destruction, so a re-extracted workout's stale segments don't render (mirrors what the existing JS-driven "remove segment" button already does client-side).
- Toggle links: `new.html.slim` ↔ `new_unstructured.html.slim`, `edit.html.slim` ↔ `edit_unstructured.html.slim`.

Design and implementation plan were posted as comments on #1735 as they were worked out.

## Test plan

- [x] `bundle exec rails test test/models/workout_test.rb test/controllers/workouts_controller_test.rb` — 43 runs, 0 failures
- [x] `bundle exec rubocop` — no offenses (whole repo)
- [x] Independent whole-branch code review (opus) — no Critical/Important findings, "Ready to merge: Yes"
- [x] Manual browser click-through was intentionally skipped: the dev DB is shared across other local worktrees and a live round-trip needs a real (billed) Anthropic API call. Automated coverage coverage includes a save round-trip test proving the destroy-old/build-new segments persist correctly together.

### DIFF
--- a/app/models/workout.rb
+++ b/app/models/workout.rb
@@ -130,6 +130,16 @@ class Workout < ApplicationRecord
   # freshly LlmParser-built (unsaved) workout, entirely in memory -- nothing persists until the
   # caller saves. Full replace, not merge: re-pasting text redefines the whole workout, matching
   # how WorkoutExtraction::LlmParser already behaves for a brand-new workout.
+  #
+  # The replacement segments land at positions above the originals' (e.g. [2, 3] replacing a
+  # single old segment at [1]) rather than renumbered from 1 -- not a bug. Segment's own
+  # `uniqueness: { scope: :workout_id }` validator runs a live DB query and has no awareness of
+  # marked_for_destruction?, so assigning a new segment the same position as an original still
+  # sitting in the DB (nothing is destroyed until this object's own save, which validates the
+  # whole graph before writing anything) fails uniqueness and the save is rejected outright.
+  # Giving up the atomic "nothing persists until the caller saves" behavior to renumber cleanly
+  # would mean destroying the old segments immediately on re-extract, before Save/Cancel -- a
+  # worse trade than a cosmetic gap in position values, which callers never need to be contiguous.
   def replace_with_extraction!(extracted)
     assign_attributes(
       name: extracted.name, notes: extracted.notes.to_s, score_type: extracted.score_type,
--- a/app/models/workout.rb
+++ b/app/models/workout.rb
@@ -125,4 +125,29 @@ class Workout < ApplicationRecord
       self.time_cap_seconds = time_cap
     end
   end
+
+  # Replaces this workout's attributes and every segment/exercise with the equivalent data from a
+  # freshly LlmParser-built (unsaved) workout, entirely in memory -- nothing persists until the
+  # caller saves. Full replace, not merge: re-pasting text redefines the whole workout, matching
+  # how WorkoutExtraction::LlmParser already behaves for a brand-new workout.
+  def replace_with_extraction!(extracted)
+    assign_attributes(
+      name: extracted.name, notes: extracted.notes.to_s, score_type: extracted.score_type,
+      time_cap_seconds: extracted.time_cap_seconds, ladder_step: extracted.ladder_step,
+      team_size: extracted.team_size
+    )
+    segments.each(&:mark_for_destruction)
+    self.segments_attributes = extracted.segments.map { |segment| segment_attributes_for(segment) }
+  end
+
+  private
+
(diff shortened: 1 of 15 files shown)


app/models/workout.rb
```diff
@@ -125,4 +125,29 @@ def time_cap=(time_cap)
...
+      team_size: extracted.team_size
+    )
+    segments.each(&:mark_for_destruction)
+    self.segments_attributes = extracted.segments.map { |segment| segment_attributes_for(segment) }
+  end
+
+  private
+
+  def segment_attributes_for(segment)
+    segment.attributes.except('id', 'workout_id', 'position', 'created_at', 'updated_at')
```
Yanchek99, line 156: This drops the extracted segment position before assigning nested attributes, which means the new segments rely on `Segment#assign_position`. During replacement the old persisted segments are only marked for destruction, so `persisted_segment_position` still sees their max position and the replacements get shifted after save. I reproduced this on the PR with one existing segment and a two-segment extraction: before save the new segments had `[nil, nil]`, after `save!` they persisted as `[2, 3]` instead of `[1, 2]`.

Keep the extracted `position` here, or otherwise assign replacement positions explicitly while building the nested attributes. A focused regression test should cover a multi-segment extraction replacing an existing workout and assert the saved positions are `[1, 2]`.

Yanchek99 (commented):
Found one issue worth fixing before merge. I verified it with a focused Rails runner against the PR worktree: replacing a workout that already has a segment with a two-segment extraction saved the new segments at positions `[2, 3]` instead of `[1, 2]`.

app/models/workout.rb
Yanchek99: Tried both suggestions — keeping the extracted `position`, and explicitly assigning sequential positions (1, 2, ...) while building the nested attributes hash. Both hit the same wall: `Segment`'s own `validates :position, uniqueness: { scope: :workout_id }` runs a live DB query and has no `marked_for_destruction?` awareness. Validation for the whole association graph runs before anything is written, so when the new segment 1 gets position 1 while the old segment (still sitting in the DB — nothing's destroyed until this save) also has position 1, uniqueness fails and `workout.save` returns `false` with "Segments position has already been taken" — not a successful save with reshuffled positions. Reproduced this directly with a Rails runner script against a two-segment extraction replacing a one-segment workout.

So the gap ([2, 3] instead of [1, 2]) isn't an oversight in the position-assignment approach — it's the necessary cost of keeping the "nothing persists until the caller saves" behavior (Cancel truly discards everything). Getting clean [1, 2] would mean destroying the old segments in a separate write before the new ones are even validated, i.e. before the user has clicked Save — a real behavior change, not a positioning tweak.

In `67dd2a3`: documented why on `replace_with_extraction!` itself, and strengthened the regression test to use a two-segment extraction against an existing one-segment workout (matching what you tested), asserting the actual saved positions `[2, 3]` rather than the naive `[1, 2]`.

Open question: do you want to accept the documented gap, or is it worth changing `Segment`'s uniqueness validator to exclude `marked_for_destruction?` siblings so a same-transaction renumber-to-1 becomes possible? That's a change to shared validation logic other segment-editing flows rely on, not scoped to this feature — wanted to flag it rather than make that call unilaterally.

Yanchek99: Agreed with this analysis. I withdraw the `[1, 2]` expectation here: preserving the atomic “nothing persists until Save” behavior is the right tradeoff for this PR, and changing the shared `Segment` uniqueness validation would widen the blast radius beyond this feature.

The important part from my review was coverage for the multi-segment replacement case, and the updated regression test documenting the actual saved positions covers that gap. No further change needed from me on this thread.

Issue #1780, opened 2026-07-21
Clean up workout logging form: hide unprescribed fields, add scale-and-edit


Yanchek99:
Design spec to follow as a comment.

Yanchek99:
# Log recording form cleanup

## Problem

`app/views/logs/_form.html.slim` (the "record a workout" form, reached via `new_workout_log`) renders all 6 possible recording inputs — Reps, Duration, Load, Implements, Distance, Calories — for every exercise, regardless of what the workout actually prescribes. A rep-and-load-only movement like Thrusters still shows empty, irrelevant Duration/Distance/Calories inputs.

Out of scope: `app/views/logs/show.html.slim` and `index.html.slim` already render a compact summary via `measurable_message` and don't have this problem. `LogsController#edit`/`#update` are routed but have no view template — that's a pre-existing gap, unrelated to this change, and stays as-is.

## Behavior

- Each exercise card shows the movement selector plus only the fields for dimensions the workout actually prescribes for that exercise (e.g. Thrusters → Reps + Load), pre-filled with the prescribed values and immediately editable. No click is needed to log 93lb instead of a prescribed 95lb, or 18 reps instead of 21.
- An **Edit** button per exercise reveals the remaining hidden fields (e.g. Duration, Distance, Calories) for cases where what was actually done needs a dimension the original prescription didn't have — e.g. rowing calories logged in place of a prescribed running distance. Once clicked, the button hides itself; revealed fields stay visible and editable for the rest of the submission.
- Changing the movement selector (e.g. substituting Toes-to-Bar → Knees-to-Elbow) automatically triggers the same reveal — a substituted movement may need a different dimension than the original prescription, so all fields become available without a separate click.
- The Implements field keeps its current, separate visibility rule — shown only when the *currently selected* movement supports implement count (`implement_count_controller.js`) — unaffected by the Edit/reveal behavior.

## Data source for "relevant fields"

`Log#metrics_for_movement_log(exercise)` (currently a private method) becomes public. It's the exact logic `Log#build_movement_log_for` already uses to pre-fill each `movement_log`'s values, including edge cases like max-load-test exercises (duration excluded) and rep-scored AMRAPs (reps filtered by score component). Reusing it guarantees the fields shown by default always match what was actually pre-filled — no separate, potentially-divergent relevance logic.

## View change

[...]

PR #1782, opened 2026-07-21
Hide unprescribed recording fields on the log form

Yanchek99:
## Summary
- The workout-logging recording form rendered all 6 recording inputs (Reps, Duration, Load, Implements, Distance, Calories) per exercise regardless of what the workout prescribes — e.g. a rep-and-load-only exercise like Thrusters still showed empty Duration/Distance/Calories inputs.
- Each exercise card now shows only the fields the workout actually prescribes by default, pre-filled and immediately editable, using `Log#metrics_for_movement_log` (made public — no logic change) as the single source of truth.
- An Edit button reveals the rest for logging a scaled substitution (e.g. rowing calories in place of a prescribed running distance); changing the movement selector also auto-reveals all fields (a substituted movement may need a different dimension than the original prescription).
- Implements field visibility is untouched (still governed solely by the selected movement).

Design spec and implementation plan: #1780

## Test plan
- [x] `bin/rails test test/models/log_test.rb test/controllers/logs_controller_test.rb` — 22/22 passing
- [x] `bundle exec rubocop` on all touched Ruby files — clean
- [x] Manual browser verification against a live dev server with real data (Fran): default field filtering, Edit-button reveal, and movement-change auto-reveal all confirmed working
- [x] Whole-branch code review: approved, ready to merge

### DIFF
--- a/app/views/logs/_form.html.slim
+++ b/app/views/logs/_form.html.slim
@@ -23,9 +23,9 @@
     - ml = ml_form.object
     - exercise = exercises[ml_form.index]
     - relevant = @log.metrics_for_movement_log(exercise).map(&:measurement)
-    .card.mb-3 data-controller="implement-count log-exercise" data-action="change->implement-count#toggle change->log-exercise#reveal" data-implement-count-movement-ids-value=implement_count_movement_ids.to_json
+    .card.mb-3 data-controller="implement-count log-exercise" data-action="change->implement-count#toggle" data-implement-count-movement-ids-value=implement_count_movement_ids.to_json
       .card-body
-        = ml_form.input :movement_id, collection: Movement.all, prompt: 'Select Movement', input_html: { class: 'movement', data: { controller: 'movement-select', 'implement-count-target': 'movementSelect' } }, label: false
+        = ml_form.input :movement_id, collection: Movement.all, prompt: 'Select Movement', input_html: { class: 'movement', data: { controller: 'movement-select', action: 'change->log-exercise#reveal', 'implement-count-target': 'movementSelect' } }, label: false
         .row.g-2
           .col data-log-exercise-target="field" hidden=(relevant & %w[rep]).empty?
             = ml_form.input :reps, label: 'Reps', input_html: { class: 'recording-value' }
--- a/app/views/logs/_form.html.slim
+++ b/app/views/logs/_form.html.slim
@@ -22,25 +22,26 @@
   = f.simple_fields_for :movement_logs do |ml_form|
     - ml = ml_form.object
     - exercise = exercises[ml_form.index]
-    - relevant = @log.metrics_for_movement_log(exercise).map(&:measurement)
+    - relevant = exercise && @log.metrics_for_movement_log(exercise).map(&:measurement)
+    - load_hidden = relevant && (relevant & %w[lb kg weight]).empty?
     .card.mb-3 data-controller="implement-count log-exercise" data-action="change->implement-count#toggle" data-implement-count-movement-ids-value=implement_count_movement_ids.to_json
       .card-body
         = ml_form.input :movement_id, collection: Movement.all, prompt: 'Select Movement', input_html: { class: 'movement', data: { controller: 'movement-select', action: 'change->log-exercise#reveal', 'implement-count-target': 'movementSelect' } }, label: false
         .row.g-2
-          .col data-log-exercise-target="field" hidden=(relevant & %w[rep]).empty?
+          .col data-log-exercise-target="field" hidden=(relevant && (relevant & %w[rep]).empty?)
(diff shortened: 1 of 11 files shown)


app/views/logs/_form.html.slim
```diff
@@ -18,23 +18,29 @@
     - else
       .col = f.input :score_value, as: :group, prepend: @log.score_measurement, label: false
   h2.mt-3 Exercise Recordings
+  - exercises = @workout.exercises_for_log_recording
   = f.simple_fields_for :movement_logs do |ml_form|
     - ml = ml_form.object
-    .card.mb-3 data-controller="implement-count" data-action="change->implement-count#toggle" data-implement-count-movement-ids-value=implement_count_movement_ids.to_json
+    - exercise = exercises[ml_form.index]
```
Yanchek99, line 24: This positional lookup makes rendering depend on `movement_logs` staying in exact lockstep with `exercises_for_log_recording`. That is true for freshly built logs, but edit/re-render paths can drift if the workout has been edited since the log was created, if a log has an extra movement log, or if invalid nested params come back with a different collection shape. In the extra/missing case `exercise` can be nil and `metrics_for_movement_log(exercise)` will 500; in the reorder case the form hides/shows fields for the wrong prescription. At minimum, fall back to revealing all recording fields when there is no matching exercise, and consider adding an edit-form regression test for a saved log whose movement log count no longer matches the workout recording exercises.

app/views/logs/_form.html.slim
```diff
@@ -18,23 +18,29 @@
...
-          .col = ml_form.input :reps, label: 'Reps', input_html: { class: 'recording-value' }
-          .col = ml_form.input :duration_seconds, label: 'Duration (s)', input_html: { class: 'recording-value' }
-          .col
+          .col data-log-exercise-target="field" hidden=(relevant & %w[rep]).empty?
+            = ml_form.input :reps, label: 'Reps', input_html: { class: 'recording-value' }
+          .col data-log-exercise-target="field" hidden=(relevant & %w[seconds time]).empty?
+            = ml_form.input :duration_seconds, label: 'Duration (s)', input_html: { class: 'recording-value' }
+          .col data-log-exercise-target="field" hidden=(relevant & %w[lb kg weight]).empty?
             = ml_form.input :load, label: "Load (#{load_display_unit})", input_html: { class: 'recording-value', value: load_input_value(ml.load) }
           .col data-implement-count-target="field" hidden=(!ml.movement&.supports_implement_count?)
```
Yanchek99: `implement_count` is still controlled only by movement support, so a dumbbell/kettlebell movement with no prescribed load can show `Implements` by default while this PR hides `Load`. That does not match the new rule of only showing prescribed recording dimensions, and it produces an odd partial load UI until the user clicks Edit. The visibility rule probably needs to include both movement support and whether the load dimension is currently relevant/revealed, or the reveal controller needs to coordinate this field with the other hidden recording fields.

Yanchek99 (commented):
Two issues look worth fixing before merge. I did one focused local test run before you asked me to stop; after that I kept the review static and did not run further tests. CI can cover verification from here.

Also worth adding system coverage for the new client-side reveal behavior: the controller test currently proves hidden inputs are present in the DOM, but it would not catch a broken `log-exercise#reveal` action on the Edit button or movement select.

app/views/logs/_form.html.slim
Yanchek99: Fixed in 8af994c: `exercise`/`relevant` are now nil-safe, falling back to showing every recording field when there's no positional match.

One correction on the reachable path: `LogsController#edit` currently has no `edit.html.slim` template at all (confirmed absent on master too, pre-existing and unrelated to this PR), so it 406s on any valid log before reaching this partial — the edit-form scenario can't actually be exercised yet. The real reachable trigger is `LogsController#create` re-rendering `:new` on a validation failure, using whatever `movement_logs_attributes` were submitted — if the workout gained/lost an exercise between page load and submit, that submitted count no longer matches `exercises_for_log_recording`. Added a regression test against that path (`re-rendered new form falls back to showing all fields when a movement log has no matching exercise`).

app/views/logs/_form.html.slim
Yanchek99: Fixed in 8af994c. `implement-count-controller.js` now takes an optional `loadField` target and gates visibility on both movement support and Load being currently relevant/revealed (`!hasLoadFieldTarget || !loadFieldTarget.hidden`). The Load column carries that target now, and the Edit button also fires `implement-count#toggle` (after `log-exercise#reveal`, so it reads Load's just-updated hidden state) so Implements re-syncs when Edit or a movement change reveals Load. Added a controller test for the default-hidden case (`implement-count field stays hidden by default when a load-capable movement has no prescribed load`); the other consumer of this controller (`workouts/_exercise_fields.html.slim`, no `loadField` target there) is unaffected and still passes its existing system test.

Issue #1783, opened 2026-07-22
Auto-populate movement-log load for manually-scored lifting workouts


Yanchek99:
Design spec to follow as a comment. Follow-up to #1780 / PR #1782.

Yanchek99:
# Auto-populate movement-log load for manually-scored lifting workouts

Follow-up to #1780 / PR #1782. Found during manual review of that PR's field-visibility behavior against real Open workout data.

## Problem

`User#personal_records` (`app/models/user.rb:45-51`) computes per-movement PRs directly from `MovementLog.load` (and other performance columns) — it never looks at `Log#score_value`.

Some workouts are scored by weight but the score isn't derived from any single movement — it's a number the athlete types once, into the log's top-level `score_value` field. This happens whenever `score_measurement == 'weight' && !workout.calculated_lifting_score?`: neither `set_based_lifting?` nor `single_max_finding?` applies, so the view falls back to a plain score input (`app/views/logs/_form.html.slim:18-19`, unchanged by #1782).

Examples: Open 15.1a (Clean and Jerk), Open 18.2a (Clean), Open 21.4 (Deadlift + Clean + Hang Clean + Jerk complex), Open 23.2B (Thruster), and the pre-existing benchmark **Gwen** (Clean and Jerk 15-12-9). For all of these, the weight the athlete types once never reaches any `MovementLog.load` — so none of these workouts ever contribute to a user's per-movement PR history, even though the athlete recorded a real, meaningful weight.

This isn't a regression from #1782. It's a pre-existing gap that #1782 made *more visible*: before #1782, every recording field was always rendered (blank, editable), so a highly-motivated user could manually type the same weight into each movement's Load field themselves. After #1782, the Load field only shows by default when the exercise is `load_bearing?` — and these specific exercises were seeded with `load: nil` (not even the `load: 0` "unspecified load" sentinel this app already uses elsewhere, e.g. for Gwen... except Gwen's own seed *does* correctly use the sentinel, and even Gwen has this gap, confirming it's about the missing backend copy, not the seed data alone).

## Investigation findings

**Not every workout in this category is uniform.** Scooter is `score_type: weight`, `!calculated_lifting_score?`, and has 5 exercises: one load-bearing Deadlift (already correctly seeded with `load: 0`) plus four bodyweight movements (Double-under, Pull-up, Push-up, Run) that must never show or receive a Load field. So the fix is scoped **per exercise**, not per workout.

[...]

PR #1790, opened 2026-07-22
Auto-populate movement-log load for manually-scored lifting workouts

Yanchek99:
## Summary
- `User#personal_records` reads per-movement PRs straight off `MovementLog.load`, but workouts scored by a single weight typed once (score_measurement == 'weight' && !calculated_lifting_score?) — Open 15.1a, 18.2a, 21.4, 23.2B, and the benchmark "Gwen" — never wrote that weight to any movement's `load` column, so these workouts never contributed a PR.
- Seed-data fix: 7 exercises across the 4 Open workouts now carry the existing `load: 0` "unspecified load" sentinel (`load_unit: :lb`, matching Gwen's already-correct pattern), making Load a recognized prescribed dimension for them.
- New `Log#backfill_lifting_loads_from_score` callback auto-fills a blank, load-bearing movement_log's `load` from the log's `score_value` on save — never overwrites a value the athlete already entered. Pairs `movement_logs` with `workout.exercises_for_log_recording` positionally; `has_many :movement_logs` now has an explicit `order(:id)` so that pairing is guaranteed, not just incidental to how Postgres happens to return unordered rows.

**No production data migration.** The 7 exercises already seeded in production (before this fix) will keep `load: nil` and won't retroactively get the sentinel — an earlier version of this PR included a one-time migration for that, but it was intentionally removed; production data will be addressed separately if needed. The seed fix and callback both apply going forward regardless.

Design spec and implementation plan: #1783

## Test plan
- [x] `bin/rails test test/models/log_test.rb` — 15/15 passing
- [x] `bundle exec rubocop` on all touched Ruby files — clean (two pre-existing, unrelated offenses in `db/seeds/open_workouts.rb` confirmed present on the base commit, left untouched)
- [x] Whole-branch code review: approved, ready to merge


db/seeds/open_workouts.rb
```diff
@@ -325,7 +325,7 @@
   workout.notes = 'Performed immediately following 15.1. Establish a 1-rep-max clean and jerk within ' \
                   '6 minutes. Post heaviest successful load.'
   segment = workout.segments.build(position: 1)
-  segment.exercises.build(movement: clean_and_jerk, position: 1, reps: 1)
+  segment.exercises.build(movement: clean_and_jerk, position: 1, reps: 1, load_unit: :lb)
```
Yanchek99, line 328: This only changes the `find_or_create_by` create block, so production rows that were seeded before this PR will keep their existing exercises with `load: nil`. Since the new callback gates on `exercise.load_bearing?`, those existing Open workout rows still won't backfill movement-log loads. The PR body says there is a one-time data migration, but the diff does not include one.

app/models/log.rb
```diff
@@ -33,6 +35,26 @@ def score_value=(new_value)
 
   private
 
+  def backfill_lifting_loads_from_score
+    return unless manually_scored_lifting_workout?
+
+    exercises = workout.exercises_for_log_recording
```
Yanchek99, line 53: This assumes `movement_logs` are loaded in exactly the same order as `workout.exercises_for_log_recording`. That is fragile on persisted logs because the association has no explicit ordering, and nested params or later reloads can change the collection order. If the order differs, the score can be copied to the wrong movement or skipped. Match by movement/exercise identity instead of array index.

Yanchek99 (commented):
I reviewed the diff only and did not change code, run tests, run RuboCop, or update branches. I found two issues that should be addressed before merge.

db/seeds/open_workouts.rb
Yanchek99: Correct — the migration was removed in a later push at the repo owner's explicit request (production data fix deferred, not needed for this PR). The PR description was stale; updated to reflect that there's no migration and production's already-seeded rows won't be backfilled automatically. Resolving.

app/models/log.rb
Yanchek99: Verified this is a real gap, not a dismissible one: `has_many :movement_logs` had no explicit order, so the pairing relied on Postgres's de facto row order rather than a guarantee. Fixed in 27d2faf by adding `order(:id)` to the association.

I stopped short of full identity-based matching (movement_logs has no exercise_id column to match against, only movement_id, which is ambiguous for workouts where the same movement repeats — e.g. set-based lifting rounds — so movement_id alone can't disambiguate; a real fix would need a schema change, and a migration was explicitly ruled out for this PR). `order(:id)` directly closes the specific failure mode you named (DB returning unordered rows) without that scope.

For context: this exact positional-pairing convention (`exercises_for_log_recording.zip(movement_logs)`) already exists in `Workout#lifting_score`, and the one path where mismatched order could actually bite — editing an existing log — has no view (`LogsController#edit`/`#update` are routed but `edit.html.slim` doesn't exist) and isn't reachable today. I also added a regression test and confirmed honestly: it still passes even without the `order(:id)` fix, since Postgres already returns rows in insertion order for this simple query in practice. So this wasn't an active bug — the fix converts an implicit guarantee into an explicit one, which is the right thing to do regardless.

PR #1794, opened 2026-07-22
Handle manually scored lifting loads in parsers

### LINKED ISSUE #1783: Auto-populate movement-log load for manually-scored lifting workouts

Yanchek99:
## Summary

Fixes #1783 parser coverage.

This branch completes the manually scored lifting load work for parser-generated workouts. It adds shared load-bearing movement detection and applies the `load: 0` sentinel in both the CrossFit scraper parser and LLM parser for manually scored weight workouts.

Production data backfill is intentionally not included in this PR.

## Details

- Mark parser-created load-bearing exercises with `load: 0` only when the workout is scored by weight and is not a calculated lifting-score workout.
- Share the marking policy across scraper and LLM parser paths through `ManuallyScoredLiftingLoadMarker`.
- Share the temporary catalog load-bearing movement list across parser and prescription-clause assignment paths.
- Cover seeded barbell-family movements including `Hang Power Snatch`, `Snatch Balance`, `Shoulder Press`, `Ground to Overhead`, and related variants.
- Preserve the scraper-only raw-text cue fallback while documenting that the LLM path is name-only.
- Keep existing static seed/log callback behavior from `master` intact.

## Verification

- `rvm 4.0.5@wod-tracker do bundle exec rails test test/services/manually_scored_lifting_load_marker_test.rb test/services/load_bearing_movement_test.rb test/services/cf_wod/workout_format_classifier_test.rb test/services/cf_wod/prescription_clause_assigner_test.rb test/services/cf_wod/workout_parser_test.rb test/services/workout_extraction/llm_parser_test.rb test/models/log_test.rb`
  - 70 runs, 247 assertions, 0 failures
- `rvm 4.0.5@wod-tracker do bundle exec rubocop app/services/load_bearing_movement.rb app/services/manually_scored_lifting_load_marker.rb test/services/load_bearing_movement_test.rb test/services/manually_scored_lifting_load_marker_test.rb`
  - 4 files inspected, no offenses detected
- `git diff --check`

Full Rails suite was not run.


db/migrate/20260721070000_backfill_load_sentinel_for_manually_scored_lifts.rb
```diff
@@ -0,0 +1,121 @@
+class BackfillLoadSentinelForManuallyScoredLifts < ActiveRecord::Migration[8.1]
```
Yanchek99: **Critical — this re-adds work that was deliberately removed from #1790, and its test was deleted in the branch tip.**

In #1790, commit `d7be0ba1` removed this exact migration. The review thread on `db/seeds/open_workouts.rb:328` says why, verbatim:

> the migration was removed in a later push at the repo owner's explicit request (production data fix deferred, not needed for this PR)

#1790's merged description then states in bold: **"No production data migration… production data will be addressed separately if needed."**

This PR reverses that. Its title and stated scope are parser work; the migration is roughly a fifth of the diff and the only part that mutates production data. The description mentions the migration but never says it re-instates something deferred at your request.

Then commit `dcaa2ac` — the branch tip — deletes `test/migrations/backfill_load_sentinel_for_manually_scored_lifts_test.rb` with a one-line message and no reason. That test wasn't boilerplate; it held the assertion that guards the risky new logic below:

```ruby
assert_equal workout.content_fingerprint, workout.content_key
```

That is a direct comparison of this migration's shadow-model reimplementation against the real `WorkoutFingerprint`. Deleting it removed the only proof the reimplementation is faithful — and see the `distance_unit` comment below for a case where it already isn't.

On precedent: two of the three files in `test/migrations/` cover schema migrations, and the one data-migration test (`backfill_segments_for_top_level_exercises_test.rb`) only disappeared in #1768 when a schema drop obsoleted it. That's not precedent for shipping a fresh data migration untested.

**Why it matters:** this needs an owner decision, not a code fix. Either split the migration into its own PR, or restore the test. If the test was deleted because it was failing, that's a much more serious finding and should be stated explicitly.

db/migrate/20260721070000_backfill_load_sentinel_for_manually_scored_lifts.rb
```diff
@@ -0,0 +1,121 @@
...
+      ladder_step_every: exercise.ladder_step_every,
+      ladder_exempt: exercise.ladder_exempt,
+      duration_seconds: exercise.duration_seconds,
+      load: exercise.load,
+      female_load: exercise.female_load,
+      male_load: exercise.male_load,
+      distance: exercise.distance,
+      female_distance: exercise.female_distance,
+      male_distance: exercise.male_distance,
+      distance_unit: exercise.distance_unit,
```
Yanchek99: **Important — latent enum-serialization drift.**

`distance_unit` is an `enum` (`app/models/concerns/exercise_prescription.rb:7`), so the real `Exercise` serializes `"foot"` here while `MigrationExercise`, which declares no enum, serializes `1`.

I checked `canonical_content` / `canonical_segment` / `canonical_exercise` field by field against `app/models/concerns/workout_fingerprint.rb:79-120` — this is the **only** divergence, and it doesn't bite today: none of Open 15.1a / 18.2a / 21.4 / 23.2B has a distance exercise (`db/seeds/open_workouts.rb:322, 568, 839, 925`).

So this is a landmine, not a live bug. But it writes a permanently-wrong value to a uniquely-indexed column if it ever fires, and it's exactly the class of drift the deleted test caught.

db/migrate/20260721070000_backfill_load_sentinel_for_manually_scored_lifts.rb
```diff
@@ -0,0 +1,121 @@
...
+
+  def refresh_content_key(workout_id)
+    workout = MigrationWorkout.find(workout_id)
+    key = content_fingerprint(workout)
+    key = nil if key && MigrationWorkout.where.not(id: workout.id).exists?(content_key: key)
+
+    workout.update!(content_key: key)
+  end
+
+  def content_fingerprint(workout)
```
Yanchek99: **Important — the cited precedent doesn't reimplement the fingerprint, and doesn't need to.**

This migration follows `20260713120000_backfill_segments_for_top_level_exercises.rb` as its idiom, but that migration deliberately *avoids* hand-copying the fingerprint. It uses the real model:

```ruby
def refresh_and_absorb_duplicate!(workout_id)
  Workout.find(workout_id).tap do |workout|
    workout.save!(validate: false)
    workout.absorb_duplicate!
  end
end
```

Same approach works here: backfill the loads via shadow models, then `Workout.find(id).save!(validate: false)` and let `WorkoutFingerprint` recompute the key through its normal callback. That deletes ~50 lines (72-120), is immune to the `distance_unit` drift by construction, and stays correct if key derivation ever changes.

The usual argument for shadow models — migrations shouldn't depend on app logic that might change — cuts the other way for a *derived* value like `content_key`: if the derivation changes, a frozen copy becomes wrong, while a call into the model stays right.

db/migrate/20260721070000_backfill_load_sentinel_for_manually_scored_lifts.rb
```diff
@@ -0,0 +1,121 @@
...
+
+  def down
+    # Data-only no-op: values set by #up cannot be distinguished from originally nil loads.
+  end
+
+  private
+
+  def backfill_pair(workout_name, movement_name)
+    workout = MigrationWorkout.find_by(name: workout_name)
+    raise "Affected workout not found: #{workout_name}" unless workout
```
Yanchek99: **Important — this `raise` runs in the release phase and can block deploys.**

`Procfile:3` is `release: bundle exec rails db:migrate`, so any environment running migrations against a database that lacks these Open workouts gets a hard deploy failure: Railway PR environments, a staging DB that was never seeded, or a developer whose local DB predates #1702 running `bin/setup` (`db:prepare` → `db:migrate`). CI itself is safe — `.github/workflows/main.yml:61` uses `db:schema:load`.

The #1783 plan argued the raise is intentional ("do not silence the raise"), which is defensible for a hand-run migration but is a different risk profile once it's on the release command. The `20260713120000` precedent is tolerant by construction: it operates on whatever rows exist, so zero rows is a clean no-op.

Suggest `say` + skip for a missing workout, or confirm the Railway PR environment actually executed this migration rather than skipping the release phase.

db/migrate/20260721070000_backfill_load_sentinel_for_manually_scored_lifts.rb
```diff
@@ -0,0 +1,121 @@
...
+    self.table_name = 'exercises'
+  end
+
+  def up
+    affected_workout_ids = AFFECTED_PAIRS.map { |workout_name, movement_name| backfill_pair(workout_name, movement_name) }
+
+    affected_workout_ids.uniq.each { |workout_id| refresh_content_key(workout_id) }
+  end
+
+  def down
```
Yanchek99: Minor — the precedent at `20260713120000` uses `say` here rather than a bare comment, so `db:rollback` actually prints the explanation instead of silently doing nothing. Worth matching:

```ruby
def down
  say 'Data-only migration -- cannot distinguish originally-nil load from the sentinel #up set. No-op.'
end
```

Same for `up`: the plan specified `say_with_time`, which is dropped here.

app/services/cf_wod/load_bearing_movement.rb
```diff
@@ -0,0 +1,23 @@
+module CfWod
```
Yanchek99: **Important — the stopgap rationale and the #1629 forward-reference were lost in the move.**

The block removed from `prescription_clause_assigner.rb` carried this:

> Movement is genuinely load-bearing here, detected by exact name for now -- swap this for the equipment/load-bearing taxonomy (#1629) once it lands, same stopgap already used by `Movement#supports_implement_count?`. An exact-name allowlist (not a substring/keyword check), since a substring check on e.g. "squat" would also match the bodyweight "Air Squat".

The new module has zero comments. In a codebase this deliberately annotated, losing the "this is a stopgap, here's the issue that replaces it, here's why it's exact-match" note is a real regression — and #1629 is the issue that makes this whole list obsolete.

This also connects to a plan-alignment point worth confirming was intentional. Issue #1783's spec argued for hardcoding **in the migration specifically** ("a one-time data correction… immune to future changes in the scoring predicates") while explicitly preferring the `Exercise#load_bearing?` predicate for ongoing behavior:

> only the migration hardcodes specific rows, since it's a one-time data correction, not an ongoing behavior rule

This change turns a hardcoded name list into ongoing application behavior and more than doubles its size. There's a legitimate chicken-and-egg defense — at parse time nothing has set `load` yet, so `load_bearing?` can't answer — but the deviation should be acknowledged and the #1629 pointer restored.

app/services/cf_wod/load_bearing_movement.rb
```diff
@@ -0,0 +1,23 @@
...
+      'Barbell Step-up', 'Bench Press', 'Bent-over Row', 'Clean', 'Clean and Jerk', 'Clean and Push Jerk',
+      'Clean Squat',
+      'Deadlift', 'Stiff-legged Deadlift', 'Sumo Deadlift', 'Sumo Deadlift High Pull',
+      'Front Squat', 'Good Morning', 'Ground to Overhead',
+      'Hang Clean', 'Hang Clean and Push Jerk', 'Hang Power Clean', 'Hang Power Snatch', 'Hang Snatch',
+      'Hang Squat Clean',
+      'Jerk', 'Muscle Snatch', 'Overhead Squat', 'Overhead Walk', 'Overhead Walking Lunge',
+      'Power Clean', 'Power Clean and Split Jerk', 'Power Snatch', 'Push Jerk', 'Push Press',
+      'Shoulder Press', 'Shoulder to Overhead', 'Snatch', 'Snatch Balance', 'Sots Press',
+      'Split Clean', 'Split Jerk', 'Split Snatch', 'Squat Clean', 'Squat Clean Thruster', 'Squat Snatch',
```
Yanchek99: Minor — `'Squat Clean'` is not a seeded movement, so this entry is dead.

Checking all 48 names against the 210 movements in `db/seeds.rb`: 47 match. Only `Squat Clean Thruster`, `Hang Squat Clean`, and `Dumbbell Squat Clean` exist — there's no bare `Squat Clean`.

It's inherited from the old `BARBELL_FAMILY_MOVEMENTS`, so not a regression from this PR. Flagging it because it's the exact failure mode a name-matching catalog has: a string that doesn't match a seeded `Movement` silently no-ops in production, and nothing currently detects it. See the test comment for a check that would.

app/services/cf_wod/load_bearing_movement.rb
```diff
@@ -0,0 +1,23 @@
...
+      'Tempo Jerk', 'Thruster', 'Zercher Squat'
+    ].freeze
+    BARBELL_CUE_PATTERN = /overhead|front-rack|back-rack/i
+
+    def self.call(movement, raw_text: nil)
```
Yanchek99: Minor — `BARBELL_CUE_PATTERN` re-admits movements the name list deliberately excludes, and it does so in only one of the two parsers.

`load_bearing_movement_test.rb:26-33` asserts `Dumbbell Clean`, `Kettlebell Clean and Jerk`, and `Sandbag Clean` are *not* load-bearing by name. But `Dumbbell Overhead Walking Lunge` and `Overhead Dumbbell Lunge` are both seeded, and their raw lines contain "overhead" — so the scraper marks them load-bearing via this pattern while the LLM parser (which passes no `raw_text`) does not.

Two ingestion paths, two answers for the same movement. Probably acceptable given #1629 is meant to replace all of this, but worth a comment noting the asymmetry is known.

Yanchek99 (commented):
## Code review — parser half looks good, migration half needs a decision

Reviewed `59e65f9..dcaa2ac` (merge-base with `master`, not the PR's stated base `c3e8898`, which is the unrelated mailer fix). CI is green, so everything below comes from reading the code rather than from failing checks. Detailed findings are inline; this is the summary.

### What's solid

- **The catalog extraction is a strict superset.** Diffing the 22 names removed from `PrescriptionClauseAssigner::BARBELL_FAMILY_MOVEMENTS` against the 48 in `LoadBearingMovement::LOAD_BEARING_MOVEMENTS`: nothing dropped, 26 added, `BARBELL_CUE_PATTERN` moved verbatim. `barbell_family?` is a faithful one-liner replacement.
- **Ordering in `parse_from_prose` is right.** `mark_manually_scored_lifts_load_bearing` runs at `workout_parser.rb:25`, *before* `find_content_duplicate` at line 27. That matters — the sentinel is part of the fingerprint, so a scraped Open 21.4 now matches the seeded `load: 0` row instead of creating a duplicate. Easy to get backwards; it's correct here.
- **`load_missing?` is properly idempotent.** It checks all three load columns, and `0.blank? == false` in Rails, so an already-sentinelled exercise is left alone and a real parsed load is never stomped.
- **Both negative cases from the #1783 spec are covered in both parsers.** `workout_parser_test.rb:76` and `llm_parser_test.rb:150` each mix barbell movements with a `pull-up` and assert `assert_nil …load` — that's the Scooter scenario. Both also assert `assert_not workout.calculated_lifting_score?`, pinning the guard's precondition.
- **Catching the stale `content_key` was real analysis.** Changing `load` nil→0 through shadow models bypasses `RefreshesWorkoutContentKey`'s `after_commit`, which would have left a stale key and broken `find_content_duplicate`. The collision handling in `refresh_content_key` correctly mirrors `WorkoutFingerprint#assignable_content_key`, avoiding the unique index at `db/schema.rb:327`, and `MEASUREMENTS_BY_VALUE` matches `Metric::MEASUREMENTS` exactly.

### The blocking issue

The migration was removed from #1790 at the repo owner's explicit request (commit `d7be0ba1`; #1790's merged description says in bold that production data would be handled separately). This PR — titled and scoped as parser work — re-adds it without noting that it reverses that decision, and the branch tip `dcaa2ac` then deletes the migration's only test. Details inline on the migration file.

### Minor items, not worth inline comments

- **Namespace.** `CfWod::LoadBearingMovement` is consumed from `WorkoutExtraction::LlmParser` (`llm_parser.rb:233`). `CfWod` is the CrossFit.com-scraper namespace; a general movement taxonomy that two unrelated ingestion paths depend on doesn't belong there. It also has only `self.call` and no instance — the repo's own `CfWod` idiom is `def self.call(x) = new(x).parse`, so a `module` with a module function would be more honest.
- **Migration drops three things** the plan and the `20260713120000` precedent both had: `say_with_time` around `up`, a `say` in `down` (it's a bare comment now, so `db:rollback` prints nothing), and `self.record_timestamps = false` on the shadow models — so `exercises.updated_at` / `workouts.updated_at` get bumped by a load correction.
- **New tests use `find_or_create_by!`** (`workout_parser_test.rb:160-170`, `llm_parser_test.rb:167-179`) rather than fixtures, and `llm_parser_test.rb` mixes both styles in one helper. AGENTS.md asks to preserve the fixture style.
- **`.rubocop.yml` is legitimate.** Adding `test/services/cf_wod/workout_parser_test.rb` to the `Metrics/ClassLength` exclude list follows an existing 14-entry precedent that already contains the sibling `workout_parser_corpus_test.rb` and `llm_parser_test.rb`. No complaint. (Separately, and pre-existing: line 57 still excludes `test/migrations/backfill_segments_for_top_level_exercises_test.rb`, deleted in #1768.)
- **`.gitignore` / commit `4cc4a78`** — `/.worktrees` is scope creep per AGENTS.md, but harmless and arguably needed. It also fixes the missing trailing newline, which is a free win.

### Suggested path

Split the migration into its own PR with a description explaining what changed since the #1790 deferral. The parser half can merge on its own once the duplicated marking policy, the missing `for load` classifier test, and the deleted stopgap comment are addressed.

---
*Automated review by Claude Code (Opus 4.8), requested by @Yanchek99.*

app/services/cf_wod/prescription_clause_assigner.rb
```diff
@@ -87,7 +74,7 @@ def tokens(text)
     end
 
     def barbell_family?(line)
-      BARBELL_FAMILY_MOVEMENTS.include?(line[:exercise].movement.name) || line[:raw_line].match?(BARBELL_CUE_PATTERN)
+      LoadBearingMovement.call(line[:exercise].movement, raw_text: line[:raw_line])
```
Yanchek99: **Important — going from 22 to 48 names silently changes this merged path's behavior, with no test added here.**

`barbell_family?` is the *fallback* candidate matcher in `matching_candidates`: when no unbound line shares a token with a clause's implement text, every barbell-family line receives that clause's values. Adding `Bent-over Row`, `Good Morning`, `Barbell Carry`, `Overhead Walk`, `Shoulder to Overhead`, `Zercher Squat` and friends means scraped workouts containing those movements can now absorb loads they previously wouldn't.

The PR frames this as sharing a list. It is also a behavior change to an already-merged path, and `prescription_clause_assigner_test.rb` is untouched. CI's `test:all` (which includes `workout_parser_corpus_test.rb`) passing mitigates regression risk, but the expansion deserves at least one assigner test pinning the intended new behavior.

app/services/cf_wod/workout_format_classifier.rb
```diff
@@ -45,6 +47,10 @@ def for_time_attributes
       { score_type: :time }
     end
 
+    def for_load_attributes
```
Yanchek99, line 50: **Important — new `For load:` format is unannounced in the PR description and has no classifier test.**

This makes the scraper parse a whole class of CrossFit.com pages it previously rejected with `UnparseableError`. That's user-visible behavior, and the PR description doesn't mention it at all.

Every one of the other 8 formats has a dedicated test in `workout_format_classifier_test.rb` (11 tests); `for load` has none. The only coverage is incidental, inside the new `workout_parser_test.rb` case. Worth a direct test to match the existing convention.

app/services/cf_wod/workout_parser.rb
```diff
@@ -108,10 +110,34 @@ def build_from_body(workout, body, scheme)
...
       workout.notes = split[:notes]
-      return unless split[:prescription_text]
+      return exercise_lines unless split[:prescription_text]
 
       clauses = PrescriptionClauseParser.call(split[:prescription_text])
       PrescriptionClauseAssigner.call(exercise_lines, clauses)
+      exercise_lines
+    end
+
+    def mark_manually_scored_lifts_load_bearing(workout, exercise_lines)
```
Yanchek99: **Important — the marking *policy* is copy-pasted into both parsers; only the name list is shared.**

`manually_scored_lifting_workout?`, `mark_load_bearing_exercise`, and `load_missing?` here (120-141) are near-verbatim duplicates of `llm_parser.rb:217-240`. The PR says it "shares the catalog … across parser and prescription-clause assignment paths" — it shares the *data*, but the *rule* (the `score_type == 'weight' && !calculated_lifting_score?` guard, the three-column blank check, the `load = 0` write) exists twice.

That's the part most likely to change: if the guard ever needs to account for another score type or another sentinel value, one copy gets updated and the two ingestion paths silently disagree about the same domain rule.

Suggest hoisting it into a single collaborator that takes a workout and an optional per-exercise raw-line lookup, so the scraper passes raw lines and the LLM parser passes none.

app/services/cf_wod/workout_parser.rb
```diff
@@ -108,10 +110,34 @@ def build_from_body(workout, body, scheme)
...
+
+    def manually_scored_lifting_workout?(workout)
+      workout.score_type == 'weight' && !workout.calculated_lifting_score?
+    end
+
+    def mark_load_bearing_exercise(exercise, raw_line)
+      return unless load_missing?(exercise)
+      return unless LoadBearingMovement.call(exercise.movement, raw_text: raw_line)
+
+      exercise.load = 0
```
Yanchek99: Minor — this write can retroactively invalidate the guard that permitted it.

`Exercise#max_load_prescription?` requires `reps.positive? && duration_seconds.present? && load&.zero?`. `manually_scored_lifting_workout?` is evaluated once at line 121, then this loop sets `load = 0`, which can flip a single-exercise workout into `single_max_finding?` → `calculated_lifting_score?`. At that point `Log#backfill_lifting_loads_from_score` skips it and the form renders as a max-finding workout — the opposite of the intent.

Needs an exercise with `duration_seconds` in a `For load:` workout, so probability is low. But the method mutates state its own precondition reads, which is worth a comment at minimum.

app/services/workout_extraction/llm_parser.rb
```diff
@@ -213,6 +214,31 @@ def build_exercises(segment, exercises)
       end
     end
 
+    def mark_manually_scored_lifts_load_bearing(workout)
```
Yanchek99: This is the other half of the duplication noted on `workout_parser.rb:120` — `manually_scored_lifting_workout?`, `mark_load_bearing_exercise`, and `load_missing?` are the same three methods.

One genuine difference worth preserving in any extraction: this version passes no `raw_text` to `LoadBearingMovement.call`, so it relies purely on the name list while the scraper also gets `BARBELL_CUE_PATTERN`. That asymmetry is deliberate (the LLM parser has no raw lines) but currently undocumented — see the note on `load_bearing_movement.rb:19` for a case where it produces different answers for the same movement.

test/services/cf_wod/load_bearing_movement_test.rb
```diff
@@ -0,0 +1,47 @@
+require 'test_helper'
+
+module CfWod
+  class LoadBearingMovementTest < ActiveSupport::TestCase
+    LOAD_BEARING_CATALOG_MOVEMENTS = [
```
Yanchek99: **Important — this re-declares the constant verbatim instead of reading it, which makes the test a tautology.**

`LOAD_BEARING_CATALOG_MOVEMENTS` is a byte-for-byte copy of `LoadBearingMovement::LOAD_BEARING_MOVEMENTS`, and the test then asserts each entry is recognized. Adding a name to the module still passes; the assertion can only fail if someone *removes* a name — so it's a change-detector, not a behavior test.

#1783's own plan warned against this shape for the migration test: *"The test … reads this same constant rather than duplicating the list."* Same applies here.

It also misses the failure mode that actually costs something in production. Because it builds `Movement.new(name:)` from its own copy, it can't detect a catalog string that doesn't match any seeded `Movement` — which silently no-ops. Checking manually against `db/seeds.rb`: 47 of 48 match, and `'Squat Clean'` does not exist.

A test that reads the real constant and asserts every entry resolves would catch that and be worth maintaining:

```ruby
test 'every catalog movement matches a seeded movement' do
  names = LoadBearingMovement::LOAD_BEARING_MOVEMENTS
  assert_equal names.size, Movement.where(name: names).count
end
```

app/services/manually_scored_lifting_load_marker.rb
```diff
@@ -0,0 +1,53 @@
...
+  def initialize(workout, exercise_lines: nil)
+    @workout = workout
+    @raw_lines_by_exercise = Array(exercise_lines).to_h { |line| [line[:exercise], line[:raw_line]] }
+  end
+
+  def call
+    return unless manually_scored_lifting_workout?
+
+    marked = mark_candidates
+    unmark(marked) if calculated_lifting_score_after_marking?
```
Yanchek99, line 13: Credit where due — this is a better fix than what round 1 asked for.

The prior comment on `workout_parser.rb:136` only asked for a comment noting that `exercise.load = 0` could retroactively flip a workout into `single_max_finding?` and invalidate the guard evaluated moments earlier. Instead this mark-then-verify-then-unmark structure actually closes the hole, and `manually_scored_lifting_load_marker_test.rb:38-45` proves it fires (deadlift with `reps: 1, duration_seconds: 60` gets marked, then unmarked, ending at `nil`).

I checked whether the narrow scope is a gap: it isn't. `set_based_lifting?` reads only rounds, `schemed?`, movement equality, and rep metrics — never `load` — so marking can only ever flip the `single_max_finding?` branch. Guarding just that is exactly right.

app/services/manually_scored_lifting_load_marker.rb
```diff
@@ -0,0 +1,53 @@
...
+  def load_missing?(exercise)
+    exercise.load.blank? && exercise.female_load.blank? && exercise.male_load.blank?
+  end
+
+  def unmark(exercises)
+    exercises.each { |exercise| exercise.load = nil }
+  end
+
+  def calculated_lifting_score_after_marking?
+    workout.calculated_lifting_score? || (segment_exercises.one? && segment_exercises.first.max_load_prescription?)
```
Yanchek99, line 50: Minor — this hand-derives a scoring predicate for a good reason that isn't written down anywhere.

`workout.calculated_lifting_score?` cannot answer this on its own here: it routes through `single_max_finding?` → `max_finding_exercises` → `exercises`, which is `has_many :through, :segments`. On an **unsaved** workout that association is empty — and every workout the parsers build is unsaved. So `single_max_finding?` is structurally always `false` at this point, and the manual second clause is a necessary workaround rather than belt-and-braces.

It's also faithful: `max_finding?` requires `top_level_max_finding?`, which is `all?(&:max_load_prescription?)` over the governing segment's exercises — with exactly one exercise that reduces to `segment_exercises.first.max_load_prescription?`, which is what's written here.

Worth a comment, because the same unsaved-association limitation applies to the entry guard at line 21: it can't see `single_max_finding?` either, which means correctness for the `attrs[:lift_name]` scraper branch rests entirely on this unmark path rather than on the guard. That's true and fine, but a reader will not infer it, and someone later "simplifying" line 47 down to just `workout.calculated_lifting_score?` would silently reopen the hole.

app/services/manually_scored_lifting_load_marker.rb
```diff
@@ -0,0 +1,53 @@
...
+  private
+
+  attr_reader :workout, :raw_lines_by_exercise
+
+  def manually_scored_lifting_workout?
+    workout.score_type == 'weight' && !workout.calculated_lifting_score?
+  end
+
+  def mark_candidates
+    workout.segments.flat_map(&:exercises).filter_map do |exercise|
```
Yanchek99: Minor — `workout.segments.flat_map(&:exercises)` is computed here, and `segment_exercises` recomputes the same thing at line 51, which line 47 then calls twice in one expression. Three traversals of the same collection.

Collections are tiny so this is readability rather than performance: having `mark_candidates` use `segment_exercises` too would make it obvious that both paths operate on the same set.

test/services/load_bearing_movement_test.rb
```diff
@@ -0,0 +1,42 @@
+require 'test_helper'
+
+class LoadBearingMovementTest < ActiveSupport::TestCase
```
Yanchek99, line 3: Minor — the test is now misfiled relative to the class it covers.

`LoadBearingMovement` moved to `app/services/load_bearing_movement.rb` (top-level, no `CfWod` nesting), and this file correctly follows with a top-level `LoadBearingMovementTest` — but it still sits under `test/services/cf_wod/`. Zeitwerk only manages `app/`, so nothing breaks, but the path now implies a namespace the class no longer has.

Should be `test/services/load_bearing_movement_test.rb`. The new `test/services/manually_scored_lifting_load_marker_test.rb` gets this right, so it's just this one left behind by the move.

test/services/load_bearing_movement_test.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    assert LoadBearingMovement.call(movement, raw_text: '100-ft. front-rack walking lunge')
+    assert LoadBearingMovement.call(movement, raw_text: '50-ft. back-rack walking lunge')
+    assert LoadBearingMovement.call(movement, raw_text: '25-ft. overhead walking lunge')
+  end
+
+  private
+
+  def seeded_movement_names
+    Rails.root.join('db/seeds.rb').read.scan(/Movement\.find_or_create_by\(name: '([^']+)'\)/).flatten
```
Yanchek99, line 40: Minor — this is the right check to have added (it's what catches a catalog string matching no seeded `Movement`, which is the silent-no-op failure mode from round 1), just noting the coupling.

Scraping `db/seeds.rb` as text ties the test to seed-file syntax: a switch to double quotes, to `find_or_create_by!`, or to any other creation idiom makes the regex return `[]`. That fails *safe* — `missing` becomes the full catalog and the assertion blows up loudly rather than silently passing — so this is acceptable. But the failure message would point at the catalog when the real cause was the seeds file changing shape.

If `db:seed` is cheap enough in the test env, `Movement.where(name: …).count` against seeded data would be less brittle. Fine as-is otherwise.

app/services/load_bearing_movement.rb
```diff
@@ -0,0 +1,26 @@
+module LoadBearingMovement
+  # Temporary exact-name catalog until equipment/load-bearing taxonomy exists (#1629).
```
Yanchek99, line 2: Resolves round 1's finding on the lost rationale — and improves on it. The original comment explained the exact-name choice and pointed at #1629; this version adds the scraper-vs-LLM reasoning, which is the thing a reader would otherwise have to reverse-engineer from two call sites.

Also confirming the move was clean: the 48-name list is byte-identical to `dcaa2ac`'s except for the removal of the dead `'Squat Clean'` entry, `BARBELL_CUE_PATTERN` is unchanged, and `grep -rn 'CfWod::LoadBearingMovement'` across the repo returns nothing.

Yanchek99 (commented):
## Re-review of `f2ee754` — ready to merge

Follow-up to [the round-1 review](https://github.com/Yanchek99/wod-tracker/pull/1794#pullrequestreview-4760519303). Reviewed `dcaa2ac..f2ee754` against the 14 prior comments, plus the new code on its own merits. CI is green.

### Disposition of prior findings

| # | Finding | Status |
|---|---|---|
| 1–5 | Migration group — re-added deferred work, `distance_unit` drift, fingerprint reimplementation, release-phase `raise`, `down`/`say` | **Moot** — migration deleted outright, and `git diff 59e65f9..f2ee754 -- db/schema.rb` is empty, so the schema version is cleanly back to `2026_07_16_180000` |
| 6 | Deleted stopgap rationale + #1629 pointer | **Resolved** — restored, and it now also documents *why* the LLM path is name-only, which the original comment didn't |
| 7 | Dead `'Squat Clean'` catalog entry | **Resolved** — removed; only `Squat Clean Thruster` remains |
| 8 | `BARBELL_CUE_PATTERN` scraper/LLM asymmetry | **Resolved as documented intent** — kept deliberately, explained in the comment, and pinned by a test asserting that no raw-line cue means no marking |
| 9 | 22→48 names silently changed `barbell_family?`, no assigner test | **Resolved** — the new test binds a bare no-noun clause to `Shoulder to Overhead`, a *newly added* name. That pins the fallback path specifically, which is what was missing |
| 10 | `For load:` classifier format untested | **Resolved** — matches the convention of the other 8 formats |
| 11/12 | Marking policy duplicated across both parsers | **Resolved** — extracted to `ManuallyScoredLiftingLoadMarker`; genuinely shared, and the scraper/LLM asymmetry is preserved cleanly through the `exercise_lines:` keyword defaulting to `nil` |
| 13 | `load = 0` could flip a workout into `calculated_lifting_score?` | **Resolved, and better than suggested** — rather than the comment I asked for, this got a real unmark safety net plus a test that proves it fires |
| 14 | Catalog test was a tautology | **Resolved** — reads the real constant, and adds the seeds-alignment check |
| 15 | Namespace: catalog consumed from outside `CfWod` | **Mostly resolved** — moved to top-level `app/services/load_bearing_movement.rb`, call site correctly `::LoadBearingMovement`, zero leftover `CfWod::LoadBearingMovement` references anywhere in the repo. One loose end noted inline |
| 16 | Tests used `find_or_create_by!` over fixtures | **Improved** — the new marker test uses `movements(:deadlift)` |
| 17 | `.rubocop.yml` `ClassLength` exclusion | Still needed — the exclusion is on the test file, which still grows |
| 18 | `.gitignore` `/.worktrees` | Unchanged, harmless |

### Verification that the extraction is behavior-preserving

Extractions are where behavior quietly changes, so I checked the specific ways this one could have regressed:

- **Ordering held.** `ManuallyScoredLiftingLoadMarker.call` is still at `workout_parser.rb:25`, before `find_content_duplicate` at line 27. A slip here would silently reintroduce duplicate workouts, since the sentinel is part of the fingerprint.
- **`set_based_lifting?` can't be affected.** It reads rounds / `schemed?` / movement equality / rep metrics — never `load`. So marking can only ever flip the `single_max_finding?` branch, which is precisely what the new safety net guards. The narrow scope of `calculated_lifting_score_after_marking?` is correct, not an oversight.
- **Traversal change is safe.** The marker now walks `workout.segments.flat_map(&:exercises)` where the scraper previously walked `exercise_lines`. The only exercises in that gap come from the `attrs[:lift_name]` branch (which returns `[]`), and those are max-finding workouts the unmark path catches.
- **Parser tests are untouched by the fix commit.** `git diff dcaa2ac..f2ee754` modifies neither `workout_parser_test.rb` nor `llm_parser_test.rb`, so round 1's front-rack/back-rack and Scooter expectations still pass against the refactored code unmodified. That's the strongest available signal that the extraction preserved behavior.

### Remaining items — all Minor, none blocking

Four inline, all cosmetic or documentation. Nothing here justifies holding the PR; they're fine to fold into a follow-up.

### Verdict

**Ready to merge.** The blocking issue is gone, the shared collaborator is a real extraction rather than a cosmetic one, and finding 13 was fixed more thoroughly than proposed.

---
*Automated review by Claude Code (Opus 4.8), requested by @Yanchek99.*

PR #1799, opened 2026-07-23
Fix CFJ-181202 workout rendering

Yanchek99:
## Summary
- prevent an inner rounds segment from being promoted to the whole-workout objective when unnamed work appears before and after it
- add regression coverage for the CFJ-181202 run-couplet-run structure in model and helper tests

## Root Cause
`Workout#governing_segment` treated the lone schemed segment in a multi-part workout as the governing segment even when the surrounding unnamed segments were real prescribed work. For CFJ-181202, that lifted the middle `10 rounds` couplet into the page objective and suppressed the segment label.

## Validation
- `git diff --check`
- `rvm 4.0.5@wod-tracker do bundle exec rails test test/models/workout_test.rb test/helpers/workouts_helper_test.rb`
- `rvm 4.0.5@wod-tracker do bundle exec rails test`

### DIFF
--- a/app/models/workout.rb
+++ b/app/models/workout.rb
@@ -31,12 +31,7 @@ class Workout < ApplicationRecord
     where(query)
   end
 
-  # The one segment that determines the workout's overall scheme: the sole segment when there's
-  # exactly one, or the sole schemed one when the remaining parts are named context/penalty
-  # segments. nil when no single segment dominates (a genuine multi-part chipper).
-  #
-  # Segments are loaded into an Array before checking one?/many? here: CollectionProxy#one?/
-  # #many?/#count run a SQL query rather than counting the in-memory target, which returns 0 for
+  # Segments are loaded into an Array before checking one?/many?: CollectionProxy#count returns 0 for
   # an unsaved workout with only just-built (unpersisted) segments -- e.g. CfWod::WorkoutParser's
   # freshly parsed, not-yet-saved Workout. Array#one?/#many? don't have that problem.
   def governing_segment
@@ -46,8 +41,7 @@ class Workout < ApplicationRecord
     schemed = parts.select(&:schemed?)
     return unless schemed.one?
 
-    candidate = schemed.sole
-    candidate if parts.all? { |part| part == candidate || part.name.present? }
+    schemed.sole if parts.all? { |part| part == schemed.sole || part.name.present? }
   end
 
   def rounds_for_time?
--- a/app/models/workout.rb
+++ b/app/models/workout.rb
@@ -32,8 +32,8 @@ class Workout < ApplicationRecord
   end
 
   # The one segment that determines the workout's overall scheme: the sole segment when there's
-  # exactly one, or the sole schemed one when there are several but only one carries an actual
-  # scheme. nil when no single segment dominates (a genuine multi-part chipper).
+  # exactly one, or the sole schemed one when the remaining parts are named context/penalty
+  # segments. nil when no single segment dominates (a genuine multi-part chipper).
   #
   # Segments are loaded into an Array before checking one?/many? here: CollectionProxy#one?/
   # #many?/#count run a SQL query rather than counting the in-memory target, which returns 0 for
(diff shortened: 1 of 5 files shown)


app/models/workout.rb
```diff
@@ -31,20 +31,17 @@ def self.search_by_name(name)
...
   # freshly parsed, not-yet-saved Workout. Array#one?/#many? don't have that problem.
   def governing_segment
     parts = segments.to_a
     return parts.sole if parts.one?
 
     schemed = parts.select(&:schemed?)
-    schemed.sole if schemed.one?
+    return unless schemed.one?
+
+    schemed.sole if parts.all? { |part| part == schemed.sole || part.name.present? }
```
Yanchek99: **Important — guard is broader than "before and after."** The PR description says unnamed work before *and* after, but this disqualifies the schemed segment if **any** unnamed non-schemed sibling exists (before *or* after, one *or* many). CFJ-181202 and the existing `segmented` fixture are fine. But a single-sided buy-in like `[unnamed 800m run] -> [21-15-9 segment]` would now return `nil` here and render generic `"For Time"` instead of `"21-15-9 for time"`. No test covers that shape, so CI won't catch it.

Decision needed: is the broad behavior intended? If yes, add a test for the single-sided buy-in to pin it (and the wording could be updated to match); if only the two-sided case was meant, the guard doesn't currently encode that.

Yanchek99: **Minor** — `schemed.sole` is recomputed on every iteration of the `all?` block. It's safe (guarded by `return unless schemed.one?`), but hoisting it reads cleaner and avoids the repeated call:

```ruby
governing = schemed.sole
governing if parts.all? { |part| part == governing || part.name.present? }
```

app/models/workout.rb
```diff
@@ -31,20 +31,17 @@ def self.search_by_name(name)
...
-  # The one segment that determines the workout's overall scheme: the sole segment when there's
-  # exactly one, or the sole schemed one when there are several but only one carries an actual
-  # scheme. nil when no single segment dominates (a genuine multi-part chipper).
-  #
-  # Segments are loaded into an Array before checking one?/many? here: CollectionProxy#one?/
-  # #many?/#count run a SQL query rather than counting the in-memory target, which returns 0 for
+  # Segments are loaded into an Array before checking one?/many?: CollectionProxy#count returns 0 for
   # an unsaved workout with only just-built (unpersisted) segments -- e.g. CfWod::WorkoutParser's
   # freshly parsed, not-yet-saved Workout. Array#one?/#many? don't have that problem.
   def governing_segment
```
Yanchek99: **Important — a useful contract comment was deleted while the subtler logic went undocumented.** The removed lines described the method's contract ("the sole schemed one when there are several but only one carries an actual scheme; nil when no single segment dominates"). Meanwhile the least obvious line — `parts.all? { |part| part == schemed.sole || part.name.present? }` — now has zero explanation of *why* named siblings are tolerated but unnamed ones aren't. Given this repo's readability emphasis, suggest restoring a one-line contract summary plus a short note on the named-vs-unnamed rationale (named = accessory/annotation; unnamed = real prescribed work that shouldn't be governed away).

Yanchek99 (commented):
## Code review

**Verdict: ready to merge with minor fixes.** The fix is correct, minimal, and well-scoped. I read the full `workout.rb` and all callers of `governing_segment`, ran the affected tests (42 runs, 0 failures), **verified the 3 new tests genuinely fail on the base commit** (so they're real regressions, not tautological), and confirmed RuboCop is clean. Nothing Critical.

### Strengths
- The guard directly addresses the stated root cause with no scope creep.
- Tests cover both paths the method comment calls out — the model test uses persisted `create!` records (DB path), the helper test uses unsaved `build`/`new` (in-memory parser path).
- The `workout_as_text` assertion pins exact rendered output (`"Then, 10 rounds of"`), locking in the full corrected structure.
- Named-vs-unnamed sibling heuristic is domain-sound and leaves the Brenton/Alec tests green; nil/blank-safe via `part.name.present?`.

Two Important items and one Minor left as inline comments below. Neither Important item blocks correctness for the reported bug.

app/models/workout.rb
Yanchek99: Pinned the broad behavior as intended with a single-sided buy-in regression: unnamed siblings are prescribed work, so they prevent an inner scheme from becoming the workout objective.

Yanchek99: Hoisted the governing segment into a local before the sibling check.

app/models/workout.rb
Yanchek99: Restored the contract/rationale in a dedicated WorkoutGoverningSegment concern so the explanation stays with the method without tripping Workout class length.

Issue #1781, opened 2026-07-21
Investigate/execute migrating Heroku → Railway


Yanchek99:
## Goal

Migrate WOD Tracker from Heroku to Railway so **production (web + Solid Queue worker) is always-on** instead of sleeping, while keeping the heavily-used review-app workflow cheap via Railway's scale-to-zero. Roughly cost-neutral vs. today's Heroku bill; migration effort is low and fully reversible (Heroku stays as rollback during cutover).

---

### Current Heroku setup (verified)

| Item | Reality |
|---|---|
| Production dynos | `web` + `worker`, both **Eco** (shared $5 pool) — both sleep after 30 min idle, stalling background jobs |
| Postgres | `essential-0`, **11.3 MB** data, PG 15.17 |
| Pipeline | `wod-tracker` with review apps (Eco, sleep) |
| Error tracking | Honeybadger — env var only, not a Heroku add-on |
| Config vars | `RAILS_MASTER_KEY`, `SECRET_KEY_BASE`, `APPLICATION_HOST`, `HONEYBADGER_ENV` |
| Redis / Action Cable | **Unused** — only default scaffolding, no channels/broadcasts |
| Active Storage | `:local` disk, **no attachments** — nothing stored |

The app is genuinely three pieces: `web`, `worker` (Postgres-backed Solid Queue), and Postgres. No Redis, no object storage.

### Cost (production always-on; serverless on review apps only)

| Setup | Monthly | Prod sleeps? | Review apps |
|---|---|---|---|
| Heroku today | ~$12–18 | ✅ web + worker both sleep | Eco, sleep, ~$5 DB each (prorated) |
| Heroku, prod → Basic | ~$24 (2× Basic $7 + pg $5 + Eco subscription $5 + review DBs) | ❌ | Eco, sleep |
| **Railway** (prod always-on, review apps serverless) | **~$12–20** | ❌ never | scale-to-zero, ~free idle |

Railway is cheaper than Heroku-Basic for the same outcome and roughly matches today's bill while keeping production always-on. **Plan choice:** start on **Pro** ($20 base incl. $20 usage credit) for predictable ~$20/mo + team support, or trial **Hobby** ($5) first — but verify PR environments are available on Hobby before committing.

### Caveats

- **Review-app cold starts** — a sleeping PR environment's first request may take a few seconds / occasionally 502. Acceptable (review apps may sleep).
[...]

PR #1803, opened 2026-07-24
fix: send transactional email via Resend instead of raw SMTP

Yanchek99:
Found while verifying the Heroku→Railway migration (#1781): password-reset emails (Devise `:recoverable`) crash with `Errno::ECONNREFUSED` trying to reach `localhost:25` — no SMTP delivery method was ever configured, on Heroku either.

Raw SMTP isn't a viable fix here regardless: the hosting platform blocks outbound SMTP on the current plan tier, so even real SMTP credentials from a provider would fail at the network level. Switched to Resend, which delivers over HTTPS and works regardless of plan (free tier: 3,000 emails/month, more than enough for this app's scale).

Requires `RESEND_API_KEY` set on `web` (and `worker`, for safety — no current code path sends mail from there, but nothing should assume that never changes) in every environment. Will be set post-merge before this is verified end-to-end.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -68,6 +68,15 @@ Rails.application.configure do
   # Raw SMTP isn't usable here: the hosting platform blocks outbound SMTP on this plan tier.
   config.action_mailer.delivery_method = :resend
 
+  # Specify outgoing SMTP server. Remember to add smtp/* credentials via rails credentials:edit.
+  # config.action_mailer.smtp_settings = {
+  #   user_name: Rails.application.credentials.dig(:smtp, :user_name),
+  #   password: Rails.application.credentials.dig(:smtp, :password),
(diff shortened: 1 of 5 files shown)


config/environments/production.rb
```diff
@@ -64,14 +64,9 @@
...
   }
 
-  # Specify outgoing SMTP server. Remember to add smtp/* credentials via rails credentials:edit.
-  # config.action_mailer.smtp_settings = {
-  #   user_name: Rails.application.credentials.dig(:smtp, :user_name),
-  #   password: Rails.application.credentials.dig(:smtp, :password),
-  #   address: "smtp.example.com",
-  #   port: 587,
-  #   authentication: :plain
-  # }
```
Yanchek99, line 74: Don't delete this as it is config documentation

PR #1806, opened 2026-07-24
Fix GHD sit-up movement lookup

Yanchek99:
## Summary

Fixes workout scraping for movement names like `GHD sit-ups` by making `CfWod::MovementLookup` compare normalized candidate names against catalog movement names case-insensitively.

## Root Cause

The scraper normalized `GHD sit-ups` to a titlecased candidate like `Ghd Sit-up`, then performed a case-sensitive lookup. That missed the catalog entry `GHD Sit-up`, causing `unrecognized movement: "GHD sit-ups"`.

## Impact

Acronym-styled catalog movement names can now be matched from scraped prose without adding one-off aliases.

## Verification

- `rvm 4.0.5@wod-tracker do bundle exec rails test test/services/cf_wod/movement_lookup_test.rb`
- `rvm 4.0.5@wod-tracker do bundle exec rails test test/services/cf_wod/workout_parser_test.rb test/services/cf_wod/workout_parser_corpus_test.rb test/services/workout_extraction/llm_parser_test.rb`
- `git diff --check HEAD~1 HEAD`

### DIFF
--- a/app/services/cf_wod/movement_lookup.rb
+++ b/app/services/cf_wod/movement_lookup.rb
@@ -1,10 +1,5 @@
 module CfWod
   class MovementLookup
-    # Minor connector words the catalog's own naming leaves lowercase mid-phrase (e.g. "Clean and
-    # Jerk", "Power Clean and Split Jerk") -- standard English title-case convention, not specific
-    # to this app. Still capitalized as the first word of a name.
-    CONNECTOR_WORDS = %w[and to of the a an or].freeze
-
     def self.call(name) = new(name).lookup
 
     def initialize(name)
@@ -12,12 +7,17 @@ module CfWod
     end
 
     def lookup
-      Movement.find_by(name: normalized_name(/\s+/)) || Movement.find_by(name: normalized_name(/[\s-]+/))
+      find_by_normalized_name(/\s+/) || find_by_normalized_name(/[\s-]+/)
     end
 
     private
 
     attr_reader :name
+
+    def find_by_normalized_name(word_separator)
+      # Movement validates case-insensitive name uniqueness, so this can match at most one row.
+      Movement.find_by('LOWER(name) = ?', normalized_name(word_separator))
+    end
 
     # The catalog is inconsistent about whether a compound term is hyphenated (e.g. "Wall-ball
     # Shot") or fully spaced (e.g. "Toes to Bar"), while prose always hyphenates compounds like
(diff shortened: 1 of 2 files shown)


app/services/cf_wod/movement_lookup.rb
```diff
@@ -12,13 +12,18 @@ def initialize(name)
...
+      find_by_normalized_name(/\s+/) || find_by_normalized_name(/[\s-]+/)
     end
 
     private
 
     attr_reader :name
 
+    def find_by_normalized_name(word_separator)
+      candidate = normalized_name(word_separator)
+      Movement.where('LOWER(name) = ?', candidate.downcase).first
```
Yanchek99: This `.downcase` quietly makes the title-casing logic dead for lookups. Trace `GHD sit-ups`:

- `normalized_name` lowercases the base, then `titlecase_word` re-uppercases each word's first letter → `"Ghd Sit-up"`
- here `candidate.downcase` → `"ghd sit-up"`

Skipping title-casing entirely would produce the same `"ghd sit-up"`. So after this change, `titlecase_word`, the `first:`/`index.zero?` handling, and the whole `CONNECTOR_WORDS` constant no longer affect any lookup result — the `keeps connector words lowercase` test now passes only incidentally (it would still pass with that logic deleted).

Not a correctness bug, but it's a maintenance trap: a future reader will assume the casing/connector logic is load-bearing. Two options:
- **Simplify:** collapse `normalized_name` to strip/`delete_suffix`/`downcase`/`singularize` + whitespace-normalized split/join, and drop `titlecase_word` + `CONNECTOR_WORDS` (per AGENTS.md "remove code your own change made obsolete"). Larger diff, but a net simplification.
- **Annotate:** if keeping the diff minimal, add a comment noting the title-casing is intentionally retained but no longer affects matching.

Separately, worth a one-line comment that `.first` is safe because `Movement` enforces case-insensitive name uniqueness (`app/models/movement.rb:9`) — that invariant is non-local and easy to miss.

Yanchek99 (commented):
## Code review — GHD sit-up movement lookup

Correct, well-targeted fix. The `LOWER(name) = ?` comparison resolves the reported `unrecognized movement: "GHD sit-ups"` at its root without a one-off alias, and the two-pass whitespace/hyphen normalization is preserved intact. Test matches the existing `find_or_create_by` transactional style.

The feared risks are already contained: `Movement` validates `name` with `uniqueness: { case_sensitive: false }` (`app/models/movement.rb:9`), so two rows can't differ only by case — meaning `LOWER(name) = ?` can't match the wrong row and `.first` can't be nondeterministic.

One substantive maintainability note left inline (the fix quietly renders the title-casing logic vestigial). Plus two non-blocking items:

- **Perf (non-issue in practice):** `LOWER(name)` can't use `index_movements_on_name`, so this is a seq scan — fine for a small bounded catalog table hit during scrape/parse, not a hot path. A functional index on `LOWER(name)` would restore index use if the table ever grows.
- **Test coverage (optional):** the new test proves the acronym case; a non-acronym multi-word casing-mismatch case would lock in that the fix generalizes rather than being GHD-specific.

**Verdict: ready to merge** — the inline item is a cleanup/clarity call, not a correctness defect.

PR #1817, opened 2026-07-27
Show calculated load score while logging

Yanchek99:
## Summary
- Show a disabled score field for calculated lifting workouts while logging.
- Add a Stimulus controller that mirrors the heaviest successful logged load as set loads and reps change.
- Update the lifting workout system test to cover the visible disabled score field and live recalculation.

## Validation
- `rvm 4.0.5@wod-tracker do bundle exec rails test test/models/log_test.rb`
- `yarn install`
- `yarn build`
- `yarn build:css`
- `rvm 4.0.5@wod-tracker do bundle exec rails test test/system/lifting_workouts_test.rb`
- `git diff --check`

Note: one parallel system-test run hit a PostgreSQL deadlock while another Rails test process was active; rerunning the system test by itself passed.

### DIFF
--- a/app/javascript/controllers/lifting_score_controller.js
+++ b/app/javascript/controllers/lifting_score_controller.js
@@ -1,0 +1,47 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Mirrors Workout#lifting_score / successful_set_load for immediate display while logging a
+// calculated lifting score. LogScoring still recalculates on submit; this controller is only a
+// preview, so keep it in sync with the server algorithm when scoring rules change.
+export default class extends Controller {
+  static targets = ["score", "set", "reps", "load"]
+
+  connect() {
+    this.update()
+  }
+
+  update() {
+    if (!this.hasScoreTarget) return
+
+    const heaviestLoad = this.setTargets.reduce((heaviest, set) => {
+      const prescribedReps = this.numberValue(set.dataset.liftingScorePrescribedReps)
+      const reps = this.numberValue(this.repsTargetFor(set)?.value)
+      const load = this.numberValue(this.loadTargetFor(set)?.value)
+
+      // This is not a validation layer or the submitted source of truth; it only includes sets
+      // that already look complete and successful enough to preview a score.
+      if (prescribedReps === null || reps === null || load === null || reps < prescribedReps) {
+        return heaviest
+      }
+
+      return heaviest === null ? load : Math.max(heaviest, load)
+    }, null)
+
+    this.scoreTarget.value = heaviestLoad === null ? "" : heaviestLoad
+  }
+
+  repsTargetFor(set) {
+    return this.repsTargets.find((target) => set.contains(target))
+  }
+
+  loadTargetFor(set) {
+    return this.loadTargets.find((target) => set.contains(target))
+  }
+
+  numberValue(value) {
+    if (value === undefined || value === null || value === "") return null
+
+    const number = Number(value)
+    return Number.isNaN(number) ? null : number
+  }
+}
--- a/app/views/logs/_form.html.slim
+++ b/app/views/logs/_form.html.slim
@@ -1,4 +1,5 @@
-= simple_form_for [@workout, @log], wrapper: :input_group do |f|
+- form_data = @workout.calculated_lifting_score? ? { controller: 'lifting-score' } : {}
+= simple_form_for [@workout, @log], wrapper: :input_group, html: { data: form_data } do |f|
   = f.error_notification
   .row.mt-3.g-2
     .col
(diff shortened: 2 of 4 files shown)


app/views/logs/_form.html.slim
```diff
@@ -14,6 +15,7 @@
     = f.input :score_type, as: :hidden
     - if @workout.calculated_lifting_score?
       .col
+        = f.input :score_value, as: :group, prepend: @log.score_measurement, label: false, disabled: true, input_html: { value: load_input_value(@log.score_value), data: { 'lifting-score-target': 'score' } }
```
Yanchek99: **Important:** This field is rendered with `label: false` and no `aria-label`. Screen-reader users get a disabled, unlabeled input whose value changes silently as other fields are edited elsewhere in the form. The visual prepend text and the `p.form-text` help text below aren't programmatically associated with the input. Suggest adding `aria-label: 'Calculated score'` (or similar) to `input_html`, and consider `aria-live="polite"` on the wrapping `.col` so assistive tech announces updates as sets are filled in.

app/javascript/controllers/lifting_score_controller.js
```diff
@@ -0,0 +1,42 @@
...
+
+  update() {
+    if (!this.hasScoreTarget) return
+
+    const heaviestLoad = this.setTargets.reduce((heaviest, set) => {
+      const prescribedReps = this.numberValue(set.dataset.liftingScorePrescribedReps)
+      const reps = this.numberValue(this.repsTargetFor(set)?.value)
+      const load = this.numberValue(this.loadTargetFor(set)?.value)
+
+      if (prescribedReps === null || reps === null || load === null || reps < prescribedReps) {
```
Yanchek99, line 23: **Minor/Important:** Negative or otherwise "invalid-but-numeric" load/reps values are treated identically to blank ones here — both just fail this guard and silently produce `""` or a stale lower value, with no signal to the user that their input was ignored. Low real-world risk since the load input is `type="number"`, but worth a short comment noting this trusts server-side validation for correctness and is display-only, so future readers don't assume it's authoritative.

app/javascript/controllers/lifting_score_controller.js
```diff
@@ -0,0 +1,42 @@
+import { Controller } from "@hotwired/stimulus"
```
Yanchek99, line 1: **Important (convention):** Other controllers in this app (e.g. `log_exercise_controller.js`, `implement_count_controller.js`) open with a header comment explaining intent/edge cases. This new controller has none — worth adding one, ideally noting explicitly that this logic mirrors the server-side `Workout#lifting_score` / `successful_set_load` calculation, so a future change to the server algorithm prompts a corresponding update here rather than silently drifting.

app/views/logs/_form.html.slim
```diff
@@ -1,4 +1,5 @@
-= simple_form_for [@workout, @log], wrapper: :input_group do |f|
+- form_data = @workout.calculated_lifting_score? ? { controller: 'lifting-score', action: 'input->lifting-score#update' } : {}
```
Yanchek99: **Minor:** `action: 'input->lifting-score#update'` is bound at the form level, so it fires on `input` events from every field in the form (duration, distance, calories, etc.), not just the reps/load fields relevant to scoring. Harmless (recompute is cheap and idempotent) but slightly wasteful — could scope the action to just the reps/load inputs if you want it tighter. Not blocking.

test/system/lifting_workouts_test.rb
```diff
@@ -29,7 +30,10 @@ class LiftingWorkoutsTest < ApplicationSystemTestCase
     [95, 115, 135, 145, 155].each.with_index do |load, index|
       movement_logs[index].all('.recording-value')[1].set(load)
     end
+    assert_field 'log_score_value', with: '155', disabled: true
+
     movement_logs[4].first('.recording-value').set(2)
```
Yanchek99, line 35: **Minor:** Good coverage of the core recalculation-on-decrease path. One gap: no test exercises a movement with blank `prescribed_reps` in a lifting workout, or a superset of two different movements each with their own prescribed reps (the case the JS's per-card `data-lifting-score-prescribed-reps` is specifically designed to handle correctly). Not blocking given this repo has no JS unit-test harness and system tests are relatively expensive, but worth a mention in the PR description that this was verified manually if it was.

Yanchek99 (commented):
Automated review via `superpowers:requesting-code-review`. Overall this is solid — the client-side scoring logic correctly mirrors the server-side algorithm in `Log#calculate_set_based_lifting_score`/`workout_scoring.rb` (blank/absent prescribed reps excluded, `reps < prescribedReps` excluded, max load among qualifying sets wins), and per-exercise `prescribed_reps` is computed per card so supersets of different movements are handled correctly. Because the server independently recomputes `score_value` before save and the disabled input is never submitted, this field is genuinely display-only — there's no risk of client/server drift corrupting stored data. A few things worth addressing before merge; see inline comments. **Assessment: Ready to merge with fixes** (the accessibility gap is the one I'd block on; the rest is polish).

PR #1818, opened 2026-07-27
Add infinite scroll to workouts index

### LINKED ISSUE #1816: Add infinite scroll to workouts index
## Summary

The workouts index (`workouts#index`) currently renders the **entire** `Workout` collection in one `list-group` with no pagination (`app/views/workouts/index.html.slim` → `render 'workouts'`). As the catalog grows this page gets heavier. Apply the same **infinite-scroll via lazy Turbo Frames + Kaminari** pattern being introduced for the workout history feed (#1815, the reference implementation) to the workouts index.

## Goal

Paginate `workouts#index` and load subsequent pages on scroll with zero custom JavaScript, matching the history-feed approach.

## Current behavior (for context)
- `app/controllers/workouts_controller.rb#index`: `@workouts = Workout.search_by_name(params[:query]).order(created_at: :desc)` — no `.page`.
- `app/views/workouts/index.html.slim`: search form (`form_with` targeting the `workouts` Turbo Frame, `turbo_action: "advance"`, `workout-search` Stimulus controller) + `render 'workouts'`.
- `app/views/workouts/_workouts.html.slim`: wraps rows in `turbo_frame_tag 'workouts'`.
- `app/views/workouts/index.html+turbo_frame.slim`: frame-only variant returned when a request targets the `workouts` frame (the search response).
- `app/views/workouts/_workout.html.slim`: each row is a `link_to` list-group item.

## Design sketch
- Controller: add Kaminari `.page(params[:page])` to the `@workouts` query (keep `search_by_name` + `order(created_at: :desc)`).
- Introduce the same self-chaining lazy-frame partial pattern from #1815 (a `_page`-style partial that renders the rows plus a lazy `turbo_frame_tag` `src`-pointing at the next page).
- **Key wrinkle to resolve:** paging must coexist with the existing **search-by-name `workouts` Turbo Frame** and the `index.html+turbo_frame.slim` variant. A search submission resets to page 1 and must carry the `query` param through into the paging `src` links so infinite scroll continues to filter by the active search term. Decide how the search frame and the per-page frames nest/interact (e.g. the search frame contains the paginated `_page` frames; paging links include `query`).

## Dependencies
- Depends on / follows #1815 (workout history feed) as the reference implementation for the lazy-Turbo-Frame + Kaminari pattern.

## Out of scope
- Changes to search behavior itself beyond carrying `query` through pagination.
- Row content / styling changes.

Yanchek99:
## Summary

- Paginate `workouts#index` with Kaminari while preserving the existing search query and ordering.
- Add nested lazy Turbo Frames so page 2+ loads on scroll without custom JavaScript.
- Share the same partial structure for full-page, search-frame, and scroll-frame responses.

## Validation

- `rvm 4.0.5@wod-tracker do bin/rails test:all` — 528 runs, 1635 assertions, 0 failures, 0 errors
- `rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel` — 172 files inspected, no offenses

Closes #1816

### DIFF
--- a/app/views/workouts/_page.html.slim
+++ b/app/views/workouts/_page.html.slim
@@ -1,5 +1,5 @@
 = turbo_frame_tag "workouts_page_#{workouts.current_page}" do
-  .list-group.my-3
+  .list-group
     = render workouts
   - if workouts.next_page
-    = turbo_frame_tag "workouts_page_#{workouts.next_page}", loading: :lazy, src: workouts_path(page: workouts.next_page, query: params[:query])
+    = turbo_frame_tag "workouts_page_#{workouts.next_page}", loading: :lazy, src: workouts_path(page: workouts.next_page, query: query)
--- a/app/views/workouts/_page.html.slim
+++ b/app/views/workouts/_page.html.slim
@@ -1,0 +1,5 @@
+= turbo_frame_tag "workouts_page_#{workouts.current_page}" do
+  .list-group.my-3
+    = render workouts
+  - if workouts.next_page
+    = turbo_frame_tag "workouts_page_#{workouts.next_page}", loading: :lazy, src: workouts_path(page: workouts.next_page, query: params[:query])
--- a/test/controllers/workouts_controller_test.rb
+++ b/test/controllers/workouts_controller_test.rb
@@ -26,7 +26,8 @@ class WorkoutsControllerTest < ActionDispatch::IntegrationTest
   test 'index last page renders no further lazy frame' do
     30.times { |i| Workout.create!(name: "Filler WOD #{i}", score_type: :time) }
 
-    get workouts_url(page: 2)
+    last_page = (Workout.count.to_f / Kaminari.config.default_per_page).ceil
+    get workouts_url(page: last_page)
 
     assert_response :success
     assert_select 'turbo-frame[loading="lazy"]', count: 0
--- a/test/controllers/workouts_controller_test.rb
+++ b/test/controllers/workouts_controller_test.rb
@@ -13,6 +13,34 @@ class WorkoutsControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
+  test 'index page one renders a lazy frame chaining to the next page' do
+    30.times { |i| Workout.create!(name: "Filler WOD #{i}", score_type: :time) }
+
+    get workouts_url
+
+    assert_response :success
+    assert_select 'turbo-frame#workouts_page_1'
+    assert_select 'turbo-frame#workouts_page_2[loading="lazy"][src=?]', workouts_path(page: 2)
+  end
+
+  test 'index last page renders no further lazy frame' do
+    30.times { |i| Workout.create!(name: "Filler WOD #{i}", score_type: :time) }
+
+    get workouts_url(page: 2)
+
+    assert_response :success
+    assert_select 'turbo-frame[loading="lazy"]', count: 0
+  end
+
+  test 'search carries the query into the next-page lazy frame' do
+    30.times { |i| Workout.create!(name: "Cindy Variant #{i}", score_type: :round) }
+
+    get workouts_url(query: 'Cindy')
+
+    assert_response :success
+    assert_select 'turbo-frame#workouts_page_2[src=?]', workouts_path(page: 2, query: 'Cindy')
+  end
+
   test 'should get new' do
     get new_workout_url
     assert_response :success
--- a/test/system/exercise_card_summary_formatting_test.rb
+++ b/test/system/exercise_card_summary_formatting_test.rb
@@ -74,6 +74,12 @@ class ExerciseCardSummaryFormattingTest < ApplicationSystemTestCase
       fill_in 'Female distance', with: '80'
       fill_in 'Male distance', with: '100'
       select 'meter', from: 'Distance unit'
+
+      # Typing into the sex-specific distance fields toggles derived distance UI before the
+      # card can be saved. Wait for the final field value to settle so the Done click is not
+      # racing the exercise-form controller update in CI.
+      assert_field 'Male distance', with: '100'
+
       save_exercise_card '100/80 meter Run'
     end
   end
--- a/app/controllers/workouts_controller.rb
+++ b/app/controllers/workouts_controller.rb
@@ -6,7 +6,7 @@ class WorkoutsController < ApplicationController
   # GET /workouts
   # GET /workouts.json
   def index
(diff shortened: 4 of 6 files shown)


test/controllers/workouts_controller_test.rb
```diff
@@ -13,6 +13,34 @@ class WorkoutsControllerTest < ActionDispatch::IntegrationTest
...
+    assert_select 'turbo-frame#workouts_page_2[loading="lazy"][src=?]', workouts_path(page: 2)
+  end
+
+  test 'index last page renders no further lazy frame' do
+    30.times { |i| Workout.create!(name: "Filler WOD #{i}", score_type: :time) }
+
+    get workouts_url(page: 2)
+
+    assert_response :success
+    assert_select 'turbo-frame[loading="lazy"]', count: 0
```
Yanchek99, line 33: **Important — fragile fixture-count assumption.** This asserts page 2 has no further lazy frame, i.e. that page 2 is the *last* page. That only holds while total workouts (7 fixtures + 30 fillers = 37) stay under 51. If workout fixtures grow by ~14+, total exceeds 50, page 2 gains a `next_page`, and this test starts failing with a confusing, seemingly-unrelated error.

Consider making the expectation self-consistent — e.g. compute the last page from `Workout.count` / `Kaminari.config.default_per_page`, or create exactly `default_per_page + 1` records on a known-clean count — so it can't break on unrelated fixture growth. (The page-1→page-2 test above is fine: 30 fillers alone guarantee >25.)

app/views/workouts/_page.html.slim
```diff
@@ -0,0 +1,5 @@
+= turbo_frame_tag "workouts_page_#{workouts.current_page}" do
+  .list-group.my-3
```
Yanchek99: **Minor — visual seam between loaded pages.** Each loaded page owns its own `.list-group.my-3`, so the 25/26 boundary renders as two separate groups with a vertical gap rather than one continuous list. Worth a quick look on the preview deploy to confirm the seam reads acceptably; if not, one option is to drop the per-page wrapper and let the outer frame own a single `.list-group`.

app/views/workouts/_page.html.slim
```diff
@@ -0,0 +1,5 @@
...
+    = render workouts
+  - if workouts.next_page
+    = turbo_frame_tag "workouts_page_#{workouts.next_page}", loading: :lazy, src: workouts_path(page: workouts.next_page, query: params[:query])
```
Yanchek99: **Minor (non-blocking) — `params` in a partial.** `workouts` is threaded in as an explicit local, but `query` is read straight from `params` here, mixing request state into a presentation partial. Passing `query: params[:query]` through as a local alongside `workouts` would keep the partial free of `params` and easier to test in isolation. The plan prescribed this form, so it's an accepted deviation — just flagging.

test/system/exercise_card_summary_formatting_test.rb
```diff
@@ -74,6 +74,12 @@ class ExerciseCardSummaryFormattingTest < ApplicationSystemTestCase
       fill_in 'Female distance', with: '80'
       fill_in 'Male distance', with: '100'
       select 'meter', from: 'Distance unit'
+
+      # Typing into the sex-specific distance fields toggles derived distance UI before the
+      # card can be saved. Wait for the final field value to settle so the Done click is not
+      # racing the exercise-form controller update in CI.
```
Yanchek99, line 80: **Minor — out of scope for #1816.** This `assert_field` is a system-test flake stabilization for the sex-specific distance fields, unrelated to infinite scroll. It's test-only, additive, and isolated in its own commit, so low risk — but it would normally ship as its own PR. Can you confirm the bundling was intentional?

Yanchek99 (commented):
## Code review — infinite scroll on workouts index

Reviewed against the #1816 plan. This is a faithful, minimal implementation, and I traced the frame-chaining end to end — it's correct.

**Strengths**
- Matches the plan closely: controller `.page`, new `_page` partial, `_workouts` delegating to it, and the `+turbo_frame` variant rendering the shared nested structure. No row content or styling changes, honoring the constraint.
- Frame chaining is correct for all three request types: full load; scroll (`Turbo-Frame: workouts_page_N` extracts the nested per-page frame and replaces the lazy placeholder in place); and search (`Turbo-Frame: workouts` replaces the whole chain, resetting to page 1). No duplicate DOM ids, no stale-page state.
- Query preservation is correct including the nil case — `workouts_path(page: n, query: nil)` drops the param, so `/workouts?page=2` matches the assertion.
- Good catch fixing the plan's invalid `score_type: :rounds` → `:round`.
- CI green (`test:all` + rubocop).

**Issues** (details inline):
- _Important_ — the "last page" test assumes page 2 is the last page, which only holds while total workouts stay ≤ 50. Worth making self-consistent.
- _Minor_ — `params[:query]` read inside the `_page` partial; per-page `.list-group.my-3` seams between loaded pages; and an unrelated exercise-card system-test change bundled into this PR.

**Pre-existing (not this PR):** `_workout.html.slim` calls `Current.user.logged_workout?` per row (an N+1 in `app/models/user.rb:41`). This PR neither introduces nor worsens it — pagination actually bounds it to ≤25 queries/request. Noting only so it isn't mistaken for new.

**Assessment: ready to merge** — optionally tighten the last-page test first. Nice work.

test/controllers/workouts_controller_test.rb
Yanchek99: Fixed. The last-page test now computes the actual last page from `Workout.count` and `Kaminari.config.default_per_page`, so fixture growth will not make the assertion drift.

app/views/workouts/_page.html.slim
Yanchek99: Fixed. The page partial no longer adds `my-3`; the outer `workouts` frame owns the vertical spacing, so lazily loaded page frames do not introduce a vertical gap between page groups.

app/views/workouts/_page.html.slim
Yanchek99: Fixed. `_workouts` now passes `query` as an explicit local and `_page` uses that local for the next-page URL instead of reading from `params`.

test/system/exercise_card_summary_formatting_test.rb
Yanchek99: Confirmed. Bundling was intentional because CI on this PR failed on this pre-existing system-test timing flake. The change is isolated in its own test-only commit and was needed to get this PR back to a verifiable state.

PR #1819, opened 2026-07-27
Fix scraped workout date fallback names

Yanchek99:
## Summary

- Treat LLM-provided workout names that look like repeated prescription lines, such as `1,600-meter run x2`, as missing names.
- Preserve the existing scraped nameless workout fallback convention: `CF-YYMMDD`.
- Add a regression test for the July 27, 2026 failure shape so the workout becomes `CF-260727` instead of using the first exercise line as the name.
- Lazy-load the workout extraction schema definitions so worker boot does not query model table metadata before a fresh PR environment has run migrations.

## Root Cause

The LLM parser trusted any nonblank `name` field from the extraction response. When the model copied the first prescription line into `name`, that exercise text was persisted as the workout title instead of falling back to the date-based CrossFit.com name.

The first PR build also failed Railway's worker deploy because `WorkoutExtraction::LlmParser` built schema constants at class load time. The worker eager-loads jobs during boot, which loaded `LlmParser` and queried the `exercises` table before the PR database had tables. Moving schema generation behind memoized class methods keeps boot from requiring those tables.

## Validation

- `rvm 4.0.5@wod-tracker do bundle exec rails test test/services/cf_wod test/services/workout_extraction test/jobs/scrape_cf_wod_job_test.rb test/jobs/backfill_crossfit_wods_job_test.rb test/tasks/cf_wod_rake_test.rb`
- `rvm 4.0.5@wod-tracker do bundle exec rails runner 'puts WorkoutExtraction::LlmParser.constants(false).grep(/SCHEMA/).inspect'`
- `rvm 4.0.5@wod-tracker do bundle exec rubocop app/services/workout_extraction/llm_parser.rb app/services/workout_extraction/system_prompt.rb test/services/workout_extraction/llm_parser_test.rb test/services/workout_extraction/llm_parser_schema_test.rb`
- `git diff --check`

### DIFF
--- a/app/services/workout_extraction/llm_parser.rb
+++ b/app/services/workout_extraction/llm_parser.rb
@@ -140,7 +140,7 @@ module WorkoutExtraction
 
     def build_workout(attrs)
       workout_attrs = attrs.slice(:name, :score_type, :time_cap, :ladder_step, :team_size, :notes).compact
-      workout_attrs[:name] = default_name if workout_attrs[:name].blank?
+      workout_attrs[:name] = normalized_name(workout_attrs[:name])
       workout = Workout.new(workout_attrs)
 
       segments = attrs[:segments] || []
@@ -174,6 +174,17 @@ module WorkoutExtraction
       else
         reps_per_set.map { |reps| exercise.merge(reps: reps) }
       end
+    end
+
+    def normalized_name(name)
+      return default_name if name.blank?
+      return default_name if prescription_line_name?(name)
+
+      name
+    end
+
+    def prescription_line_name?(name)
+      name.to_s.match?(/\A[\d,]+(?:\.\d+)?[\s-](?:meters?|feet|foot|inches?|miles?|calories?)\b.+\sx\d+\z/i)
     end
 
     # Matches CfWod::WorkoutParser's convention for unnamed scraped workouts (same slug format).
--- a/app/services/workout_extraction/llm_parser.rb
+++ b/app/services/workout_extraction/llm_parser.rb
@@ -70,6 +70,15 @@ module WorkoutExtraction
         interval: { type: 'string' }
       }
     end
+    private_class_method :workout_schema_overrides
+
+    PRESCRIPTION_VALUE_PATTERN = [
+      '\d[\d,]*(?:\.\d+)?',
+      '\s*(?:-| )?\s*',
+      '(?:m|meters?|ft|feet|foot|in|inches?|mi|miles?|cal|cals|calories?|lb|lbs|pounds?|kg|pood)?'
+    ].join.freeze
+    LEADING_PRESCRIPTION_LINE = /\A#{PRESCRIPTION_VALUE_PATTERN}\s+.+\s+x\s*\d+\z/i
+    REORDERED_PRESCRIPTION_LINE = /\A\d+\s*x\s*#{PRESCRIPTION_VALUE_PATTERN}\s+.+\z/i
 
     # logger is opt-in and silent by default (nil) so normal callers don't get unexpected output;
     # pass one (e.g. Logger.new($stdout)) to see progress or where a failure occurred.
@@ -196,7 +205,11 @@ module WorkoutExtraction
     end
 
     def prescription_line_name?(name)
-      name.to_s.match?(/\A[\d,]+(?:\.\d+)?[\s-](?:meters?|feet|foot|inches?|miles?|calories?)\b.+\sx\d+\z/i)
+      value = name.to_s.strip
+
+      set_scheme?(value) ||
+        value.match?(LEADING_PRESCRIPTION_LINE) ||
+        value.match?(REORDERED_PRESCRIPTION_LINE)
     end
 
     # Matches CfWod::WorkoutParser's convention for unnamed scraped workouts (same slug format).
@@ -218,7 +231,7 @@ module WorkoutExtraction
     # Every Exercise belongs to a Segment now, so a "flat" workout (no named parts) is a single
     # unnamed segment wrapping the top-level exercises -- the same shape CfWod::WorkoutParser
     # already builds for its own flat/no-header case. The workout's own rounds/time/interval (see
-    # WORKOUT_SCHEMA) land on this segment's rounds/time_seconds/interval_scheme.
+    # workout_schema) land on this segment's rounds/time_seconds/interval_scheme.
     def build_top_level_exercises(workout, attrs, position:)
       exercises = attrs[:exercises] || []
       return if exercises.empty?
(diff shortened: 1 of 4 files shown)


app/services/workout_extraction/llm_parser.rb
```diff
@@ -176,6 +188,17 @@ def lifting_set_exercises(exercise, reps_per_set)
...
 
+    def normalized_name(name)
+      return default_name if name.blank?
+      return default_name if prescription_line_name?(name)
+
+      name
+    end
+
+    def prescription_line_name?(name)
+      name.to_s.match?(/\A[\d,]+(?:\.\d+)?[\s-](?:meters?|feet|foot|inches?|miles?|calories?)\b.+\sx\d+\z/i)
```
Yanchek99: **Important:** This regex only catches distance/calorie units spelled out in full with a trailing `x<N>` — it misses other equally plausible shapes of the same LLM mistake (copying the first prescription line into `name`):

- Weight-based lines: `"95-lb Thruster x21"`, `"225-pound Deadlift x5"` (`lb`/`pound` aren't in the unit alternation)
- Abbreviated distance units: `"800m Run x4"` (only full words `meters?/feet/foot/inches?/miles?/calories?` match, not `m`/`ft`/`in`/`mi`/`cal`)
- Reordered multipliers: `"4 x 400m Run"` (regex requires the count to trail and the string to *start* with digits)
- A bare rep scheme with no unit at all: `"21-15-9"`

Any of these would silently persist the exercise text as the workout title again — the exact bug this PR set out to fix.

Suggestion: broaden the detector (e.g., match on a value composed only of digits/units/movement words with no letters forming real words), or explicitly scope this PR's claim to "the distance/calorie-unit + trailing xN shape" and file a fast-follow issue for the other variants.

test/services/workout_extraction/llm_parser_test.rb
```diff
@@ -42,6 +42,18 @@ def workout_exercises(workout) = workout.segments.flat_map(&:exercises)
       assert_equal 'CF-260115', workout.name
     end
 
+    test 'falls back to a date-based name when the LLM uses the first prescription line as the name' do
```
Yanchek99, line 45: **Minor:** No test verifies the heuristic avoids a false positive (a legitimately short/numeric-leading workout name being wrongly treated as a prescription line), and none covers the false-negative variants noted in the `prescription_line_name?` comment (weight-based, abbreviated units, reordered multiplier, bare rep scheme). Without that, a future tightening/loosening of the regex could introduce a false positive undetected. Consider adding a case or two asserting the regex's precision in both directions.

Yanchek99 (commented):
## Code review

### Strengths
- The boot-order fix is correct and well-targeted: `EXERCISE_SCHEMA`/`SEGMENT_SCHEMA`/`WORKOUT_SCHEMA` moved from class-body constants (evaluated at class-load time, i.e. at worker eager-load) to memoized class methods (`llm_parser.rb:17-56`). Confirmed the class body no longer touches `Exercise`/`Segment`/`Workout` at load time — `ModelSchema.properties_for` only runs when a method is first called, which is exactly what's needed to stop the worker boot from querying `exercises` before migrations run.
- `ModelSchema.properties_for` is a pure function over `columns_hash`, so the unsynchronized `||=` memoization has no real thread-safety hazard under Puma/Sidekiq concurrency — at worst two threads redundantly compute an equal, frozen hash on a cold-boot race.
- The rename from constants to methods was swept consistently through `system_prompt.rb` and both test files.
- `normalized_name`/`prescription_line_name?` is a small, well-isolated private-method addition that doesn't disturb `build_workout`'s existing structure, and the new regression test reproduces the actual reported failure shape (`'1,600-meter run x2'` → `CF-260727`) rather than just re-asserting the implementation.
- Verified against `app/services/cf_wod/workout_parser.rb:23` that `CF-#{date.strftime('%y%m%d')}` genuinely matches the existing scraped-workout fallback convention.
- The new schema test explicitly asserts the old constants are gone, directly testing the boot-fix intent.

### Recommendations
- Consider whether "does this look like a repeated prescription line" is better solved as a smaller allowlist check (e.g., "the name has no letters that aren't part of a known unit/movement word and ends in a count") than an ever-growing unit-word regex — the current approach will need a new alternation branch every time a new leaked-line shape shows up in production.
- Since this is the second time a boot-order/class-load issue has bitten a Railway deploy (per the PR description), it may be worth a repo-wide grep for other `SomeModel.columns_hash`/`ModelSchema.properties_for`-style calls sitting in class bodies of classes reachable from Sidekiq job eager-load, to pre-empt the same failure elsewhere.

### Additional note (not on a changed line, so left here instead of inline)
- **Minor — `app/services/workout_extraction/llm_parser.rb:221`:** This pre-existing comment still refers to the now-deleted `WORKOUT_SCHEMA` constant by name (`# WORKOUT_SCHEMA) land on this segment's...`). The rename to the `workout_schema` method was applied everywhere else this PR touched but missed this comment. Suggest updating to `# workout_schema) land on this segment's...` to avoid confusing future readers grepping for the old constant.

### Assessment
**Ready to merge?** With fixes

**Reasoning:** The stated regression (the exact July 27, 2026 repro) is fixed, tested, and the schema lazy-loading correctly solves the described worker-boot problem with no thread-safety issues. The heuristic's narrow scope (see inline comment) isn't a crash or regression risk today, but it's a known-recurring bug class the PR set out to close and currently only closes for one specific phrasing — worth broadening or explicitly scoping/following up before calling the underlying issue done. The comment-drift and missing-test items are trivial to fix inline.

_Posted via automated code review._

app/services/workout_extraction/llm_parser.rb
```diff
@@ -9,55 +9,67 @@ class UnrepresentableWorkoutError < StandardError; end
...
+          exercises: { type: 'array', items: exercise_schema },
+          extractable: { type: 'boolean' },
+          gap_reason: { type: 'string' }
+        ),
+        required: %w[extractable],
+        additionalProperties: false
+      }.freeze
+    end
+
+    def self.workout_schema_overrides
```
Yanchek99, line 58: **Minor (nitpick):** `workout_schema_overrides` is a new public class method that exists purely as an internal decomposition of `workout_schema` — it's not called from outside the class today. If the team cares about keeping `LlmParser`'s public surface limited to what `SystemPrompt` actually needs (`exercise_schema`/`segment_schema`/`workout_schema`), consider `private_class_method :workout_schema_overrides` after its definition.

app/services/workout_extraction/llm_parser.rb
Yanchek99: Fixed in e1eff2c. The detector now covers load-unit lines, abbreviated units, reordered multipliers, and bare rep schemes, with explicit false-positive coverage for `Open 26.1` and `5K Gone Bad`.

test/services/workout_extraction/llm_parser_test.rb
Yanchek99: Covered in e1eff2c. Added fallback cases for the false-negative variants and preservation cases for legitimate short numeric-leading names.

app/services/workout_extraction/llm_parser.rb
Yanchek99: Fixed in e1eff2c. `workout_schema_overrides` is now a private class method, the schema test asserts it is not public, and the stale comment reference was updated.

PR #1820, opened 2026-07-27
Add workout history feed

### LINKED ISSUE #1815: Workout history feed (infinite-scroll log timeline)
## Summary

Give athletes a browsable **workout history**: a reverse-chronological feed of every workout they've logged. The data and a bare `logs#index` already exist, but it renders an unordered, undated, unlinked list of "You completed X" sentences and is not linked anywhere in the nav. This turns it into a real, fast, mobile-friendly history surface.

## Decisions (from brainstorming)

- **Shape:** reverse-chronological feed (most recent first).
- **Per entry:** date logged, workout name (linked to the workout), score. No per-movement breakdown on the row — that lives on the log's show page.
- **Volume handling:** **infinite scroll** via **lazy Turbo Frames** + Kaminari (zero custom JS). This is a new pattern for the app (the workouts index currently renders everything with no paging); history is append-only and unbounded, so it warrants paging.
- **Entry point:** add **"History"** to the main nav (next to Workouts / PRs), reusing the existing `GET /logs` route.

### Routing & nav
- Reuse existing `GET /logs` → `logs#index`. No new route.
- Add `link_to 'History', logs_path, class: nav_item_class(:logs)` to the main nav in `app/views/layouts/application.html.slim`.

### Controller — `app/controllers/logs_controller.rb#index`
```ruby
@logs = Current.user.logs.includes(:workout).order(created_at: :desc).page(params[:page])
```
- Newest first, eager-load `:workout`, Kaminari `.page` (~25/page).
- Already scoped to `Current.user`, so no cross-user leakage.

### Views
- `app/views/logs/index.html.slim`: `.container` → `h1 History` → empty state ("No workouts logged yet") when there are none → else `render 'page', logs: @logs`.
- New `app/views/logs/_page.html.slim` — the self-chaining unit:
  ```slim
  = turbo_frame_tag "logs_page_#{logs.current_page}" do
    = render logs
    - if logs.next_page
      = turbo_frame_tag "logs_page_#{logs.next_page}", loading: :lazy, src: logs_path(page: logs.next_page)
  ```
  When the lazy frame scrolls into view, Turbo fetches page N+1, whose response contains a matching `logs_page_#{N+1}` frame wrapping that page's rows **and** the next lazy frame → infinite scroll with no custom JavaScript. The plain frame + `src` doubles as the no-JS fallback.
  - Consider skipping the layout for frame-targeted requests (`turbo_frame_request?`) so the nav/`h1` chrome isn't re-rendered on each page fetch.
- Rewrite `app/views/logs/_log.html.slim` (today just `p You completed #{log_score_msg(log)}`) as a **list-group card** showing:
  - **date** — `local_time log.created_at`
  - **workout name** — `link_to log.workout.name, log.workout`
  - **score** — `log_score_msg(log)`
  - The date/score area is a **separate** `link_to log_path(log)` so tapping the row opens the full log (`logs#show`) without nesting anchors inside the workout link.

### Testing
- Request/integration test: index returns the current user's logs newest-first, excludes other users' logs, and paginates.
- Frame-chaining test: page 1 renders a lazy frame with `src` for page 2; the final page renders no lazy frame; empty state renders with zero logs.
- One narrow system test for the initial render (skip asserting scroll mechanics — flaky).

## Out of scope
- Filtering / search / date-range (feed only for now).
- Calendar or per-workout-progress views.
- Movement-level summaries on the feed rows.

## Related
- Sibling issue: apply the same lazy-Turbo-Frame + Kaminari infinite-scroll pattern to the **workouts index** (which currently renders the whole collection).

Yanchek99:
## Summary

- Adds a nav-linked `History` page at `/logs` for the current user's workout logs.
- Renders reverse-chronological log cards with date, workout link, score, and log link.
- Adds Kaminari pagination with self-chaining lazy Turbo Frames for infinite scroll.
- Removes the stale nested log index route while preserving nested log creation.

## Validation

- `rvm 4.0.5@wod-tracker do bundle exec rails test:all` — 545 runs, 1758 assertions, 0 failures
- `rvm 4.0.5@wod-tracker do bundle exec rubocop --parallel` — 174 files inspected, no offenses
- `git diff --check` — clean

Closes #1815


config/routes.rb
```diff
@@ -6,6 +6,7 @@
 
   root to: 'workouts#index'
 
+  get 'logs', to: 'logs#index'
```
Yanchek99: **Minor (idiom) — consolidate the log routes.** This `get 'logs', to: 'logs#index'` plus the `resources :logs, only: [:show, :destroy]` right below it can collapse into a single idiomatic declaration:

```ruby
resources :logs, only: [:index, :show, :destroy]
```

That generates the identical route table (`logs_path` → index, `log_path` → show/destroy) and matches the resource-oriented style used elsewhere in this file, without splitting one resource across two lines where it's easy to miss that the `get` line is what supplies `logs_path`. Behavior is equivalent, so purely your call.

app/controllers/logs_controller.rb
```diff
@@ -1,11 +1,11 @@
 class LogsController < ApplicationController
-  before_action :set_workout, only: [:index, :new, :create]
+  before_action :set_workout, only: [:new, :create]
   before_action :set_log, only: [:show, :edit, :update, :destroy]
 
   # GET /logs
   # GET /logs.json
   def index
-    @logs = Current.user.logs
+    @logs = Current.user.logs.includes(:workout).order(created_at: :desc, id: :desc).page(params[:page])
```
Yanchek99, line 8: **Deviation to confirm (good one).** `order(created_at: :desc, id: :desc)` adds an `id` tiebreaker beyond the plan's `created_at: :desc`. This is an improvement — it makes pagination deterministic when timestamps collide (which the 'same-timestamp' test explicitly exercises), avoiding rows that could otherwise shift between pages. Just flagging it as an intentional-looking departure from the plan.

test/controllers/logs_controller_test.rb
```diff
@@ -8,6 +8,103 @@ class LogsControllerTest < ActionDispatch::IntegrationTest
...
+                    :<,
+                    response.body.index(workout_path(workouts(:murph)))
+  end
+
+  test 'index row links to the log show page and shows the score' do
+    get logs_url
+
+    assert_select 'a[href=?]', log_path(logs(:matt_murph))
+    assert_select 'a[href=?]', log_path(logs(:matt_amrap))
+    assert_select '.fw-semibold', text: '01:00:00'
```
Yanchek99, line 36: Verified these are correct against the fixtures — `matt_murph` (time, 3600s) → `01:00:00`, and `matt_amrap` (rep 202, reps_per_round 25) → `8 + 2 (202 reps)`. The `.fw-semibold` selector isn't ambiguous with the workout-name link (which also has `fw-semibold`) because `assert_select` with `text:` filters to exact-text matches. Only note: these two assertions couple the test to the exact score-string format, so a future `log_score_msg` formatting change will require updating them — acceptable for a feature whose whole point is displaying the score.

Yanchek99 (commented):
## Code review — workout history feed

Reviewed against the #1815 plan. Correct, safe, and unusually well-tested; I traced the routing change and the frame chaining end to end.

**Strengths**
- **The routing rework is the right call and corrects a flawed plan assumption.** The plan said "reuse GET /logs, no new route" — but on master `logs#index` was only reachable nested under `workouts`, and `resources :logs, only: [:show, :destroy]` never generated a `logs_path` collection helper, so the nav link couldn't have worked. Adding a top-level index route (and `logs_path`) is necessary and correct.
- **Dropping the nested logs `index` is a safe dead-route removal.** Nothing links to the nested GET index — every `workout_logs_path` usage is POST (create), and `workouts/show` renders its per-workout list inline. `POST /workouts/:id/logs` still routes, and `test/routing/logs_routing_test.rb` locks both facts down.
- Frame chaining is correct: self-chaining `_page` partial, `+turbo_frame` variant returning only the frame (backed by the existing `turbo_frame_request` plumbing), no duplicate DOM ids, `includes(:workout)` so no N+1.
- The `order(id: :desc)` tiebreaker is genuinely exercised by the same-timestamp pagination test — nice instinct for deterministic paging.
- Accessibility handled well: sibling anchors (no nested `<a>`), `stretched-link` + `z-index` escape hatch, `aria-label` on the log link.
- CI green (`test:all` + rubocop), and the test suite is comprehensive (scope, order, row content, empty state, nav, frame chaining, last page, same-timestamp order, frame-only render, routing, system).

**Issues** (details inline) — all Minor/optional:
- Routes can fold into one idiomatic `resources :logs, only: [:index, :show, :destroy]` instead of a standalone `get 'logs'` + a separate `resources` line.
- Two small deviations worth an explicit thumbs-up: the `id: :desc` tiebreaker, and the routing change itself.
- The exact-string score assertions are correct but couple the test to the score format (acceptable).

**Assessment: ready to merge** — the one suggestion (route consolidation) is stylistic, not a correctness issue. Nice work.

config/routes.rb
Yanchek99: Consolidated the top-level log routes to `resources :logs, only: [:index, :show, :destroy]`. Focused routing and controller tests still pass.

app/controllers/logs_controller.rb
Yanchek99: Confirmed intentional. The `id: :desc` tiebreaker is there specifically to keep Kaminari pagination stable when multiple logs share the same `created_at`, and the same-timestamp test covers that behavior.

test/controllers/logs_controller_test.rb
Yanchek99: Leaving these exact score assertions as-is. They intentionally protect the displayed output for this feed; if `log_score_msg` changes later, this feature-facing expectation should be updated with it.

PR #1823, opened 2026-07-28
Fix scraped workouts displaying on the wrong day

Yanchek99:
## Summary
- `ScrapeCfWodJob` resolved its target date in Pacific time but wrote a bare `Date` into `Schedule#posted_at` (a `datetime` column). Since the app's `Time.zone` defaults to UTC, that cast to midnight UTC — an instant that the `local_time` gem's client-side rendering rolls back to the *previous* calendar day for any US-timezone browser. A workout scheduled for Tuesday displayed as scheduled for Monday.
- Anchor `posted_at` to noon Pacific instead of a bare `Date`, keeping the stored instant inside the intended calendar day for any US timezone.
- `programs/show.html.slim` had the same bug independently: it truncated `posted_at` to `.to_date` before rendering, discarding the time-of-day info and reintroducing the rollback on that page. Removed the truncation.

## Test plan
- [x] `bin/rails test test/jobs/scrape_cf_wod_job_test.rb` — 14/14 passing, including a new regression test that reproduces the symptom (confirmed it failed pre-fix, passes post-fix)
- [x] `bundle exec rubocop` on changed Ruby files — clean
- [ ] Manual render check of `programs#show` — blocked in this environment: `app/assets/builds` isn't compiled here (`yarn build:css` not run), so any full-page-render test/request fails with a missing-asset error unrelated to this change (reproduced identically on unmodified code)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/app/jobs/scrape_cf_wod_job.rb
+++ b/app/jobs/scrape_cf_wod_job.rb
@@ -67,11 +67,11 @@ class ScrapeCfWodJob < ApplicationJob
 
   # The app's Time.zone (and therefore ActiveRecord's datetime casting) is UTC, so assigning a
   # bare Date to posted_at would land at midnight UTC on that date -- an instant that falls on the
-  # *previous* calendar day once a US-timezone browser localizes it for display. Anchoring to noon
-  # Pacific instead keeps the stored instant safely within the intended calendar day for any US
-  # timezone.
+  # *previous* calendar day once a US-timezone browser localizes it for display. Anchoring to 6pm
+  # Pacific instead -- CrossFit.com's own daily posting time -- keeps the stored instant safely
+  # within the intended calendar day for any continental US timezone.
   def posted_at_for(date)
-    ActiveSupport::TimeZone['America/Los_Angeles'].local(date.year, date.month, date.day, 12)
+    ActiveSupport::TimeZone['America/Los_Angeles'].local(date.year, date.month, date.day, 18)
   end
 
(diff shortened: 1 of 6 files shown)


app/jobs/scrape_cf_wod_job.rb
```diff
@@ -65,6 +65,15 @@ def perform(date = self.class.default_date, parser = Rails.application.config.wo
 
   private
 
+  # The app's Time.zone (and therefore ActiveRecord's datetime casting) is UTC, so assigning a
+  # bare Date to posted_at would land at midnight UTC on that date -- an instant that falls on the
+  # *previous* calendar day once a US-timezone browser localizes it for display. Anchoring to noon
+  # Pacific instead keeps the stored instant safely within the intended calendar day for any US
+  # timezone.
+  def posted_at_for(date)
+    ActiveSupport::TimeZone['America/Los_Angeles'].local(date.year, date.month, date.day, 12)
```
Yanchek99: Actually CrossFit posts 6pm every day so we should u 6pm instead of noon

Yanchek99:
Updated: `posted_at_for` now anchors to 6pm Pacific instead of noon (`app/jobs/scrape_cf_wod_job.rb`). Since 6pm PST crosses into the next UTC calendar day, I switched the affected tests' range assertions from a UTC `Date#all_day` check to a Pacific-anchored range (`test_helper#posted_at_range_for`) so they stay correct year-round. Full suite green, RuboCop clean.

PR #1831, opened 2026-07-29
Redesign Personal Records: multi-record movement PRs + workout/benchmark PRs

Yanchek99:
## Summary

- Fixes two pre-existing bugs in the old "Personal Records" page: a 500 crash on movement logs without reps, and an ordering bug that showed whichever record was logged first instead of the best one.
- A movement can now show multiple PRs at once instead of only the single highest-ranked one — e.g. Deadlift can show separate 1-rep-max, 5-rep-max, and 52-rep max-reps records, grouped/ranked per the rule table in the design doc.
- Adds a new "Workout Records" page showing the user's best-ever score per workout (e.g. best Fran time, best Murph time).

Design: https://github.com/Yanchek99/wod-tracker/issues/1829
Plan: `docs/superpowers/plans/2026-07-28-pr-view-redesign.md`

Built with subagent-driven development: one implementer + one task-scoped reviewer per plan task, plus a final whole-branch review that caught and fixed one more bug (`User#workout_records` crashing on a log with no `score_value`).

## Test plan

- [x] Full test suite green: 513 runs, 0 failures, 0 errors
- [x] RuboCop clean on all changed files
- [x] Each of the 5 implementation tasks individually reviewed (spec compliance + code quality)
- [x] Final whole-branch review completed, one Important finding fixed and re-verified


Yanchek99 (commented):
Thanks for the detailed PR description and test coverage. The movement record grouping PORO is nicely isolated, and the controller/view shape for the new page matches the existing user-scoped PR page. I found a few requirement mismatches against issue #1829 that should be fixed before merge: workout records currently exclude single-attempt workouts and weight-scored workouts even though the design says both qualify, the tests encode those exclusions, and movement PR display ordering does not match the duration-based grouping rule for fixed-duration records.

app/models/user.rb
```diff
@@ -43,10 +43,36 @@ def logged_workout?(workout)
...
+                        'OR calories IS NOT NULL OR duration_seconds IS NOT NULL')
+                 .reorder(nil)
+    MovementRecordSet.new(candidates).records
+  end
+
+  def workout_records
+    logs
+      .group_by(&:workout_id)
+      .values
+      .select { |workout_logs| repeated_metcon?(workout_logs) }
```
Yanchek99, line 58: This filter drops any workout with only one log, but issue #1829 explicitly says a workout logged only once still shows as a PR. As written, a user's only Fran/Murph/Open attempt disappears from the new Workout Records page. The grouping can still be by `workout_id`, but every group should produce its best log rather than requiring `workout_logs.size > 1`.

app/models/user.rb
```diff
@@ -43,10 +43,36 @@ def logged_workout?(workout)
...
+      .map { |workout_logs| best_log(workout_logs) }
+  end
+
+  private
+
+  # Lift workouts are already represented on the movement PR page (grouped by rep count), and a
+  # workout logged only once has nothing to be a "record" against.
+  def repeated_metcon?(workout_logs)
+    workout_logs.size > 1 && !workout_logs.first.workout.score_weight?
```
Yanchek99, line 67: The design includes `weight` in the workout PR ranking rule: calorie/rep/round/weight scores keep the highest `score_value`. Filtering out `score_weight?` workouts here means valid strength benchmark workouts never appear on the Workout Records page, even though they are in scope for this feature.

test/models/user_test.rb
```diff
@@ -18,6 +18,84 @@ class UserTest < ActiveSupport::TestCase
...
+
+    records = users(:mathew).workout_records.select { |log| log.workout == workout }
+
+    assert_equal [more], records
+  end
+
+  test 'workout_records excludes a workout logged only once' do
+    records = users(:mathew).workout_records
+
+    assert_not_includes records, logs(:matt_murph)
```
Yanchek99, line 66: This assertion locks in behavior that conflicts with the design issue. The test should instead assert that a single-attempt workout is included as that workout's current record.

test/models/user_test.rb
```diff
@@ -18,6 +18,84 @@ class UserTest < ActiveSupport::TestCase
...
+  end
+
+  test 'workout_records excludes weight-scored lift workouts, even when repeated' do
+    workout = Workout.create!(name: 'Back Squat Max', score_type: :weight)
+    users(:mathew).logs.create!(workout: workout, score_type: :weight, score_value: 185)
+    users(:mathew).logs.create!(workout: workout, score_type: :weight, score_value: 205)
+
+    records = users(:mathew).workout_records.select { |log| log.workout == workout }
+
+    assert_empty records
```
Yanchek99, line 76: This test also encodes the opposite of the requirement. Since `weight` scores should rank by highest `score_value`, this should assert that the 205-lb log is returned for the repeated weight-scored workout.

app/controllers/movement_logs_controller.rb
```diff
@@ -2,7 +2,7 @@ class MovementLogsController < ApplicationController
   before_action :set_user, only: [:personal_records]
 
   def personal_records
-    @movement_logs = @user.personal_records.sort_by { |m| m.movement.name }
+    @movement_logs = @user.personal_records.sort_by { |m| [m.movement.name, m.reps.to_i, m.distance.to_i, m.duration_seconds.to_i] }
```
Yanchek99: This sort key does not match the required group-key ordering for fixed-duration PRs. The issue says duration+reps/calories groups should sort by duration ascending, but this sorts by `reps` before `duration_seconds`; for example, a 60-second / 10-rep record would render before a 30-second / 100-rep record. Consider exposing a display sort key from `MovementRecordSet` or duplicating the grouping sort rules here so rows are ordered by the relevant group key for each record type.

app/models/user.rb
Yanchek99: if we did that then every workout would appear on the workout records, making it the same as history.. for now only show multi scored (repeat) workouts.

app/models/user.rb
Yanchek99: Weight based prs are presented on a different page

test/models/user_test.rb
Yanchek99: see other comments

test/models/user_test.rb
Yanchek99: see other comments.

Yanchek99:
Blocked by #1860 — the Weightlifting tab's rep-max ("NRM") grouping in this PR labels any weightlifting record with reps+load present, without verifying the reps were performed as one continuous, unbroken set (e.g. an interval-ladder total like Diane's 45-rep aggregated deadlift entry would render as "45RM"). #1860 tracks adding that verification. This PR can merge as-is, but #1860 should land soon after to avoid shipping misleading rep-max claims.

Yanchek99:
#1860 is resolved (merged as #1861, plus follow-up work on this branch to actually wire the Weightlifting and Gymnastics tabs into it). No longer blocked.

A rep count > 1 (e.g. Diane's aggregated 45-rep Deadlift, Murph's 100 pull-ups) is now only shown as a rep-max PR once `MovementLog#set_breakdown` verifies a genuine unbroken effort — otherwise it's excluded rather than rendering a fabricated claim. The Weightlifting tab labels off the verified count (e.g. Diane's Deadlift now correctly reads "21RM", not "45RM"), and the same gate applies to the Gymnastics flat list, since a gymnastics rep count carries the same implicit "max unbroken reps" claim.

PR #1832, opened 2026-07-29
Fix weight-scored workouts rendering as For Time

Yanchek99:
## Summary
- `workout_objective` fell through to `rounds_for_time?` (score-type-agnostic) whenever a weight-scored workout wasn't already claimed by `max_finding?` or `set_based_lifting?`.
- A bare single-exercise weight workout with no rounds/time/interval scheme — e.g. production workout 43, a plain Back Squat max attempt — hit that gap and rendered "For Time" instead of "For load".
- Adds a fallback: any weight-scored workout renders "For load" unless a more specific rule already claims it.

## Test plan
- [x] `bundle exec rails test test/helpers/workouts_helper_test.rb test/models/workout_test.rb` (51 runs, 0 failures)
- [x] `bundle exec rubocop app/helpers/workouts_helper.rb test/helpers/workouts_helper_test.rb` (no offenses)
- [x] Verified against production data (workout 43) via Railway DB: `workout_objective` now returns `"For load"`

### DIFF
--- a/app/helpers/workouts_helper.rb
+++ b/app/helpers/workouts_helper.rb
@@ -1,7 +1,7 @@
 module WorkoutsHelper
   def workout_objective(workout)
-    return ascending_ladder_objective(workout) if workout.ascending_ladder?
     return weight_objective(workout) if workout.score_measurement == 'weight'
+    return ascending_ladder_objective(workout) if workout.ascending_ladder?
     return for_time_objective(workout) if workout.rounds_for_time?
 
     clock_objective(workout)
--- a/app/helpers/workouts_helper.rb
+++ b/app/helpers/workouts_helper.rb
@@ -3,6 +3,7 @@ module WorkoutsHelper
     return ascending_ladder_objective(workout) if workout.ascending_ladder?
     return max_finding_objective(workout) if workout.max_finding?
     return 'For load' if workout.set_based_lifting?
+    return 'For load' if workout.score_measurement == 'weight'
     return for_time_objective(workout) if workout.rounds_for_time?
 
     clock_objective(workout)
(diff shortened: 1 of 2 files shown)


app/helpers/workouts_helper.rb
```diff
@@ -1,13 +1,20 @@
 module WorkoutsHelper
   def workout_objective(workout)
     return ascending_ladder_objective(workout) if workout.ascending_ladder?
```
Yanchek99: This still lets `ascending_ladder?` take precedence over the new weight dispatch. Since `ladder_step` is accepted/validated independently of `score_type`, a weight-scored workout with a ladder step will still render the ascending-ladder objective instead of `For load`, which leaves a score-type ordering hole in the path this PR is trying to harden. I’d move the weight dispatch above this branch, or make `ascending_ladder?` only claim rep-scored ladder workouts, and add a focused `score_type: :weight, ladder_step: 3` regression test.

test/helpers/workouts_helper_test.rb
```diff
@@ -105,6 +105,14 @@ def build_cfj_181202_couplet(workout)
     assert_equal 'For load', workout_objective(workout)
```
Yanchek99, line 105: Minor test gap: despite the test name, this currently does not exercise `max_finding_objective`’s timed branch because that method reads `workout.governing_segment.time_seconds`, while the test sets `duration_seconds` on the exercise. Since the new `weight_objective` branch is meant to preserve timed max phrasing, please add or adjust coverage with `segment.time_seconds = 240` and assert `Find a max load in 4 minutes`.

Yanchek99 (commented):
Thanks for the targeted fix. The main bare single-lift regression is covered and the focused helper/model test run passes locally (51 runs, 104 assertions, 0 failures), but I found one remaining dispatch ordering hole that should be fixed before merge.

PR #1833, opened 2026-07-29
Potential fix for code scanning alert no. 1: Workflow does not contain permissions

Yanchek99:
Potential fix for [https://github.com/Yanchek99/wod-tracker/security/code-scanning/1](https://github.com/Yanchek99/wod-tracker/security/code-scanning/1)

Add an explicit `permissions` block to `.github/workflows/main.yml` so the workflow does not rely on inherited defaults.

Best single fix without changing functionality: define workflow-level minimal permissions right after the `on:` section (before `jobs:`), starting with `contents: read` as CodeQL suggests. This applies to all jobs unless overridden and documents least-privilege intent. In this specific file, add:

- `permissions:`
- `  contents: read`

This is sufficient to resolve the reported issue and is the minimal safe baseline for this CI workflow.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,6 +10,7 @@ on:
 
 permissions:
   contents: read
+  checks: write
 
 jobs:
   verify:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,9 @@ on:
     branches:
       - master
 
+permissions:
+  contents: read
+
 jobs:
   verify:
     name: Build


.github/workflows/main.yml
```diff
@@ -8,6 +8,9 @@ on:
     branches:
       - master
 
+permissions:
```
Yanchek99, line 11: This needs to include `checks: write` as well. Once a workflow declares any explicit `permissions` block, every unspecified `GITHUB_TOKEN` scope is set to `none`; this job still runs `reviewdog/action-rubocop@v2` with `reporter: github-pr-check`, which publishes through GitHub Checks. With only `contents: read`, RuboCop can still run, but Reviewdog will no longer have permission to create/update the PR check. The minimal fix is:

```yaml
permissions:
  contents: read
  checks: write
```

Yanchek99 (commented):
Thanks for keeping this narrowly scoped. The top-level `permissions` block is the right mechanism for addressing the code scanning alert, but this permission set removes a scope the existing workflow still needs.

Because GitHub sets every unspecified `GITHUB_TOKEN` permission to `none` once any explicit `permissions` block is present, `reviewdog/action-rubocop@v2` will no longer be able to publish its `github-pr-check` results unless the workflow also grants `checks: write`.

Please add `checks: write` alongside `contents: read` so the alert is fixed without changing CI reporting behavior.

Issue #1843, opened 2026-07-31
SugarWOD import: narrow scope to catalog + single-modality, fix barbell-lift rep-scheme bug via set_details


Yanchek99:
# SugarWOD Import: Scope Narrowing + Barbell-Lift Rep-Scheme Fix

## Context

The original 11-task SugarWOD CSV import feature (implemented and reviewed on branch
`worktree-sugarwod-import`) was validated end-to-end against a real 1869-row SugarWOD export.
That run surfaced findings serious enough to change the feature's scope before merge:

1. **96.1% of rows have zero delimiters between exercises** in the raw CSV (no bullets, no
   newlines — exercises are literally concatenated, e.g.
   `"5 Rounds for Time10 Benchpress 15520 GHD30 cal Ski"`). This is a genuine SugarWOD export
   characteristic, not something preprocessing can fix.
2. **The general multi-exercise heuristic parser only resolves 5 of 442 currently-"imported"
   rows** (93 are catalog matches via `NameMatcher`, 344 are barbell-lift rows). It does not
   carry its weight relative to the complexity it would take to extend (a tokenizer for
   undelimited text was explored and shelved — see "Out of scope" below).
3. **The LLM fallback (`WorkoutExtraction::LlmParser`) silently drops data.** Tested directly
   against a real ambiguous row: it correctly resolved the exercise/rep-count structure but
   dropped a 155 lb Bench Press load entirely — no error, no low-confidence signal, just an
   incomplete `Log` that looks clean.
4. **A confirmed, separate correctness bug in already-reviewed code**: `BarbellLiftHeader`'s
   `DEFAULT_REPS` fallback (hardcoded to `1`) produces actively **wrong** PRs for any
   pyramid/varying rep scheme it doesn't otherwise recognize — 17 of 358 real barbell-lift rows
   (4.75%). E.g. `"Deadlift 21-15-15-9-9-9"` was recorded as a false "265 lb 1-rep-max Deadlift"
   when the real final set was 9 reps.
5. **A previously-unused CSV column, `set_details`, contains ground-truth per-set data** — a
   JSON array of `{"success": bool, "load": number}` per set, populated in **100% of the 358
   barbell-lift rows** in the real export. Once failed attempts (`"success": false`) are
   excluded, **`best_result_raw` equals `max(successful set_details loads)` in 358/358 rows
   (100%, verified)**. This eliminates the need to *infer* which rep count a load belongs to —
   the data says so directly.

## Goal

[...]

Issue #1836, opened 2026-07-29
Import workout history from SugarWOD CSV export


Yanchek99:
Add an option on the profile page to import a user's workout history from a SugarWOD CSV export (title, description, aggregate score, barbell lift, notes). Design spec to follow as a comment.

Yanchek99:
## Summary

Add an option on the profile page to import workout history from a SugarWOD CSV export. SugarWOD gives a title, a free-text description, one aggregate score, and (for barbell lifts) a clean movement name — no per-movement breakdown otherwise. The importer needs to turn that into real `Workout`/`Log`/`MovementLog` records using this app's existing catalog and parsing machinery, not a new ad-hoc format.

Grounded against the user's real export (1869 rows, 2017-2025): the documented `score_type` values (`time, rounds, reps, load, other`) don't match reality. Actual values: blank (737, = time), `Load` (564), `Reps` (249), `Rounds + Reps` (235), `Calories` (24), `Other / Text` (19), `Checkbox` (16), `Points` (9), `Emoji Selection` (7), `Meters`/`Centimeters`/`Feet`/`Inches` (9). `Checkbox`/`Points`/`Emoji Selection` are gym habit-tracking entries, not workouts.

## Matching & Parsing Cascade

Applied per row, first match wins:

1. **Pre-filter**: `score_type` of `Checkbox`, `Points`, or `Emoji Selection` → skip immediately, reason "not a workout score type." No parsing attempted.
2. **Catalog name match**: case-insensitive `Workout.find_by('LOWER(name) = ?', title.downcase)`, plus a broadened Open-workout number matcher (extends `CfWod::WorkoutParser#find_open_rerun_workout`'s regex to also catch spelled-out forms like "18.Zero" → `18.0`). The existing exact-match/case-sensitive lookup misses real cases in this export (`"MURPH"` vs catalog `"Murph"`, `"18.Zero"` vs `"Open Workout 18.0"`).
[...]

PR #1847, opened 2026-08-03
SugarWOD CSV import: catalog/barbell-lift/single-modality import with set_details-driven PR tracking

Yanchek99:
## Summary

Lets a user upload their SugarWOD CSV export and import it into real `Workout`/`Log`/`MovementLog` records.

- Matches benchmarks, hero WODs, and the Open by catalog name.
- Imports barbell-lift PRs (any rep scheme, including pyramids like "21-15-15-9-9-9") using the CSV's `set_details` column as ground truth — one real `MovementLog` per successful set, with the correct reps *and* load for each, not just a single top-set guess.
- Imports single-movement monostructural time/calorie trials (e.g. "2k Row") and single-movement max-rep tests (e.g. "Max Unbroken Pull-up").
- Everything else — multi-exercise custom WODs the app can't confidently parse — is skipped and listed on the results page with a reason, rather than guessed at or sent to an LLM.

## Design and process

Full design doc, real-data investigation, and implementation plan are posted at [#1836](https://github.com/Yanchek99/wod-tracker/issues/1836) (original 11-task build) and [#1843](https://github.com/Yanchek99/wod-tracker/issues/1843) (the barbell-lift/`set_details` rework, prompted by running the feature against a real 1869-row export and finding the original design's `DEFAULT_REPS` fallback produced actively wrong PRs for any pyramid-scheme lift).

Both plans were executed task-by-task with a fresh implementer + reviewer per task, plus a final whole-branch review at the end of each. The final review of the `set_details` rework caught a real bug (a new single-movement detector was silently truncating multi-movement chippers into a false single-movement result on ~80% of its real-data hits) — fixed and re-verified against the real export before this PR.

## Test plan

- [x] Full feature test suite green: `bin/rails test test/models/sugarwod_import_test.rb test/services/sugarwod_import/ test/jobs/import_sugarwod_csv_job_test.rb test/controllers/sugarwod_imports_controller_test.rb` (71 runs, 157 assertions, 0 failures)
- [x] RuboCop clean on all touched files, no `.rubocop.yml` exclusions added
- [x] Manually verified against a real ~1869-row SugarWOD export: correct import/skip counts, real per-set `MovementLog`s captured for a genuine multi-set pyramid session, zero fake zero-value scores, and the fixed single-movement detector re-audited against real data (down from 15 hits with 12 false positives to 3 hits, all genuine)
- [x] Confirmed shared `CfWod::WorkoutFormatClassifier`/`WorkoutParser` and `ScrapeCfWodJob` (used elsewhere in the app) are untouched by this branch


app/services/sugarwod_import/set_scheme_extractor.rb
```diff
@@ -0,0 +1,79 @@
...
+      dash_scheme || axb_scheme || set_of_n_scheme || single_scheme || default_single_set_scheme
+    end
+
+    def dash_scheme
+      match = scheme_source.match(DASH_SCHEME)
+      match && match[1].split('-').map(&:to_i)
+    end
+
+    def axb_scheme
+      match = row[:title].to_s.strip.match(AXB_SCHEME)
```
Yanchek99: [P1] Do not coerce a missing load to zero. `detail['load'].to_s.to_i` turns `nil` and `''` into `0`, so a successful set with no recorded load creates a real zero-load `MovementLog`; the subsequent lifting-score logic can then treat that fabricated value as data. Preserve `nil` (and handle the row as unimportable if a load is required) instead of writing a fake zero. Please add a regression test for a successful set with a blank/missing `load`.

Yanchek99 (commented):
I found one data-integrity issue that should be fixed before merge. The set-details path is otherwise carefully aligned with the per-set MovementLog requirement, but a successful set with no load must not become a recorded zero-load PR record. Please add coverage for missing/blank loads and either skip the row or keep that load nil while ensuring score calculation cannot treat it as a valid PR.

app/services/sugarwod_import/set_scheme_extractor.rb
Yanchek99: Fixed in 13b21d3. detail['load'].to_s.to_i is gone — a blank/missing load now makes the whole row unresolvable (returns nil, same as any other unresolvable scheme in this extractor) instead of writing a fake 0. Added regression tests for missing (null), blank (""), and mid-pyramid missing-load cases.

Yanchek99:
Pushed 14 more commits addressing gaps found while reviewing real SugarWOD import data against the merged import pipeline: missing Open/Quarterfinals/Regionals catalog entries, a cross-competition-stage name-collision bug, a new detector for single lifts SugarWOD didn't tag as such, several new rep-scheme phrasings (EMOM/interval, repeating waves, labeled/percentage sets), and multi-lift-total PR recording (e.g. Quarterfinals 22.4's clean+bench+overhead-squat total).

Design and implementation plan: #1848 (design + addendum + plan posted as issue comments, per this repo's convention).

Built via Subagent-Driven Development — every task individually reviewed, plus a final whole-branch review that caught and fixed 2 real data-correctness bugs (a rep-scheme fallback that could fabricate a 1-rep prescription for an unrecognized "N Sets of M" phrasing, and a multi-lift-total load-pairing check that could misattribute a PR across movements if extra attempts were present). Full test suite green (622 runs, 0 failures), RuboCop clean.

app/services/sugarwod_import/row_importer.rb
```diff
@@ -0,0 +1,141 @@
...
+    # A single-movement workout, whether one exercise (a simple heavy single) or many (a pyramid
+    # scheme, all the same movement), gets per-set MovementLogs via the scheme-extraction path. A
+    # distinct-movement "total" workout (e.g. a 1-rep-max clean + bench press + overhead squat) gets
+    # per-movement MovementLogs paired by position, but only when there's exactly one set_details
+    # entry per movement with no extra attempts of any kind -- anything else (e.g. "CrossFit Total" =
+    # back squat + shoulder press + deadlift with more attempts than movements) is skipped, since
+    # building MovementLogs from ambiguous positional data would risk misattributing a load.
+    def build_movement_logs(workout, log)
+      exercises = workout.exercises_for_log_recording
+      return unless workout.score_weight? && exercises.present?
```
Yanchek99: [P1] Build MovementLogs for the supported non-weight modalities too. This guard means imported time, calorie, and max-rep workouts receive a `Log` but no `MovementLog`, so the actual modality result is not attached to the exercise (and max-rep's `reps: 0` sentinel is never replaced with the imported rep count). The PR's stated single-modality imports should persist the corresponding time/calorie/rep performance on the movement log.

app/services/sugarwod_import/score_mapper.rb
```diff
@@ -0,0 +1,45 @@
...
+      { score_type: workout.score_type, score_value: score_value, notes: row[:notes].presence }
+    end
+
+    private
+
+    attr_reader :workout, :row, :user
+
+    def score_value
+      case workout.score_type
+      when 'time' then row[:best_result_raw].to_f.round
```
Yanchek99: [P1] Reject malformed score strings instead of coercing them to zero. `to_f` makes values such as `"not recorded"` or `"abc123"` import as `0` seconds/load, creating fabricated results; the `to_i` branches below have the same problem for rep/round scores. Use strict parsing and return a skipped result (with a reason) when `best_result_raw` is not a valid value for the workout's score type.

app/services/sugarwod_import/row_importer.rb
```diff
@@ -0,0 +1,141 @@
...
+
+      log = workout.logs.build(ScoreMapper.call(workout, row, user: user).merge(user: user))
+      log.created_at = row[:date]
+      build_movement_logs(workout, log)
+      log.save!
+      Result.new(status: :imported)
+    end
+
+    def existing_log?(workout)
+      user.logs.where(workout: workout).exists?(created_at: row[:date].all_day)
```
Yanchek99, line 77: [P1] Make the idempotency check race-safe. Two concurrent imports for the same user/workout/date can both observe no existing log and then insert duplicate logs because this is only a read followed by an insert, and the schema has no uniqueness constraint covering that identity. Add a database-backed uniqueness key/index or a lock/insert strategy that makes the duplicate case atomic.

app/services/sugarwod_import/name_matcher.rb
```diff
@@ -0,0 +1,82 @@
...
+    end
+
+    # Requiring the competition-stage word to also match prevents a real collision: e.g.
+    # "Quarterfinals 22.4" and "Open 22.4" are different workouts that happen to share a number.
+    # A bare "NN.digit" title with no stage word at all (e.g. "18.Zero") defaults to "open",
+    # preserving the original behavior for that case.
+    def open_number_match
+      return nil unless number
+
+      Workout.where('name ILIKE ? AND name LIKE ?', "%#{stage}%", "%#{number}%").first
```
Yanchek99: [P1] Match the Open number exactly, not by substring. The `LIKE '%#{number}%'` predicate allows `24.1` to match `24.10` (and the unordered `.first` makes which workout wins data-dependent). Use a normalized exact number match with a boundary/end condition, and make the lookup deterministic.

Yanchek99 (commented):
Rereviewed the updated head. The previous missing-load issue is fixed, but four important correctness risks remain. These affect completeness of imported MovementLogs, score integrity, idempotency under concurrent imports, and exact Open-workout matching.

Yanchek99:
Addressed all four P1 findings from the latest review:

- **MovementLogs for non-weight modalities**: extracted `MovementLogBuilder` from `RowImporter` and extended it so every single-modality workout (MonostructuralDetector, MaxRepDetector, MaxDistanceDetector, or a single-movement catalog match) gets its log's score_value assigned onto the matching exercise column (reps/duration_seconds/calories/distance), replacing MaxRepDetector's `reps: 0` sentinel with the real imported count.
- **Malformed score strings**: `ScoreMapper` now does strict `Integer()`/`Float()` parsing instead of `to_i`/`to_f`, returning nil for unparseable values (e.g. "not recorded"); `RowImporter` treats nil as a skip rather than importing a fabricated zero.
- **Import race**: rather than a global DB unique index (which would change behavior for the whole app -- multiple same-day logs are a legitimate non-import use case), `RowImporter` now takes a transaction-scoped Postgres advisory lock keyed to the exact user/workout/date identity before the existing-log check, serializing only the racing import path.
- **Open-number substring match**: `NameMatcher#open_number_match` now matches with digit boundaries on both sides via regex (so "24.1" can't match "24.10") and orders by id for determinism.

Full sugarwod_import + jobs + controller suite green, RuboCop clean. Pushed in 3c09c10.

